PR #65, opened 2016-07-28
Finish mockup

carlvitzthum:
Finished (barring any other suggestions) the 1.0 mockup


j1z0:
best I can tell saucelabs was just being temperamental, re-ran and got all the tests to pass.

PR #116, opened 2016-08-10
79 80 analysis

aschroed:
Resolving issues 79 and 80 this will be ready to merge into master once the error in rendering the workflow due to something going wrong in auditor.py?  Likely has something to do with the 'analysis_steps' property and object(s) within.


j1z0:
@aschroed - I pushed the fix for this, you were trying to embed analysis_steps, but its a property of workflow not a separate item (and you only embed separate items).

It's more confusing because your workflow has a property called analysis_steps and there is also and analysis_step object that has the collection name analysis_steps... to avoid confusion please consider renaming the workflow property to something like workflow steps or something.

aschroed:
Ok - this should now be ready to merge with master.

aschroed:
Merge done.

Issue #130, opened 2016-08-12
Move enums for experiment_type to specific classes of experiment


aschroed:
The required type of experiment should be constrained by an enum list that is specific for that class of experiment.  For example, experiment_HiC can be of type 'DNase I Hi-C', 'in situ Hi-C' ... but not of type 'confocal imaging' or 'CHIA-PET' which will be types that are allowed in their specific schemas.

PR #131, opened 2016-08-12
Exp type

aschroed:
#130 fixed.


j1z0:
looking good @aschroed

PR #186, opened 2016-08-28
Added experimentset_published

KorayKirli:


aschroed:
What do you think about renaming this field to something that reflects what these are a little better i.e. the experiment sets that are described in this publication?  Fully written out it would be something like 'publication_experiment_sets' but we could shorten that to either 'pub_experiment_sets' or 'publication_exp_sets' or at the most extreme 'pub_exp_sets'.

PR #204, opened 2016-09-09
React upgrade, including redux and react-bootstrap implementation

carlvitzthum:
@j1z0 please review


src/encoded/static/browser.js
```diff
@@ -2,11 +2,25 @@
...
 require('./libs/react-patches');
 var React = require('react');
+var ReactDOM = require('react-dom')
 var ReactMount = require('react/lib/ReactMount');
 ReactMount.allowFullPageRender = true;
 
 var App = require('./components');
 var domready = require('domready');
+var store = require('./store');
+import { Provider, connect } from 'react-redux';
```
j1z0: we have require and import, which after babel gets finished with them both end up as the same code (which is actually require) anyways I think we should standardise on using one for consistencies sake.

src/encoded/static/components/app.js
```diff
@@ -235,23 +237,35 @@ var App = React.createClass({
             banners.push(<home.BannerLoader text='experiments' location='/search/?type=Experiment&award.project=4DN'/>);
             banners.push(<home.BannerLoader text='experiments' location='/search/?type=Experiment&award.project=External'/>);
             banners.push(<home.BannerLoader text='cell types' location='/search/?type=Biosource'/>);
+            // banners.push(<div> -- </div>);
+            // banners.push(<div> -- </div>);
+            // banners.push(<div> -- </div>);
```
j1z0: can we remove unnecessary commented lines

src/encoded/static/components/app.js
```diff
@@ -235,23 +237,35 @@ var App = React.createClass({
...
+                    title = title + ' – ' + portal.portal_title;
+                } else {
+                    title = portal.portal_title;
+                }
+            }else{
+                // This code is here to deal with non-simultaneous changing
+                // of props.href and props.context. This content below should
+                // never be displayed.
+                content = <ErrorPage />;
+                title="Not Found";
```
j1z0: if the error is for non-simultaneous changing of props.href vs props.context should the tile / error page say that as opposed to Not Found?

src/encoded/static/components/app.js
```diff
@@ -235,23 +237,35 @@ var App = React.createClass({
...
             }
         }
         // Google does not update the content of 301 redirected pages
         var base;
         if (({'http://www.encodeproject.org/': 1, 'http://encodeproject.org/': 1})[canonical]) {
```
j1z0: when we go live the url will be http://data.4dnucleome.org/.  Can you adjust accordingly please?

src/encoded/static/components/audit.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 var React = require('react');
 var _ = require('underscore');
-var {Panel} = require('../libs/bootstrap/panel');
+import { Panel } from 'react-bootstrap';
```
j1z0: same as previous comment please standardize on require or import throughout entire project.

src/encoded/static/components/item.js
```diff
@@ -6,39 +6,41 @@ var globals = require('./globals');
 var audit = require('./audit');
 var form = require('./form');
 var _ = require('underscore');
-var cx = require('react/lib/cx');
-var panel = require('../libs/bootstrap/panel');
+import { Panel } from 'react-bootstrap';
```
j1z0: require vs import

src/encoded/static/components/item.js
```diff
@@ -6,39 +6,41 @@ var globals = require('./globals');
...
-                </header>
-                {typeof context.description == "string" ? <p className="description">{context.description}</p> : null}
-                <section className="view-detail panel">
-                    <div className="container">
-                        <pre>{JSON.stringify(context, null, 4)}</pre>
-                    </div>
-                </section>
-            </div>
-        );
+    // render: function() {
```
j1z0: the nice thing about this error page vs the basic error page that was put in that just says page not found is that this error page provides a bit more details about what the error might actually be.  I would thus for for keeping the "old style" error page in so we get more descriptive info.

src/encoded/static/components/item.js
```diff
@@ -148,7 +146,7 @@ globals.listing_titles.fallback = function () {
     return title;
 };
 
-
+// Removed this functionality for now...
```
j1z0: just a note on style, adding a TODO at the start of comments that imply more action to do in the future is helpful so that the code base can be grepped to determine outstanding technical debt that needs to be taken care of.

src/encoded/static/components/login.js
```diff
@@ -1,13 +1,11 @@
 'use strict';
 var React = require('react');
-var Modal = require('react-bootstrap/lib/Modal');
-var OverlayMixin = require('react-bootstrap/lib/OverlayMixin');
+import { Modal } from 'react-bootstrap';
```
j1z0: import vs require

src/encoded/static/components/login.js
```diff
@@ -27,33 +25,26 @@ var Login = React.createClass({
         } else { //if logged in give them a logout link
             userActionRender = <a href="#" onClick={this.handleToggle}  data-trigger="logout" className="global-entry">Sign out</a>;
         }
-        return (<div>{userActionRender}</div>);
+        return (
```
j1z0: perhaps add an issue to address this later, but we should probably changing the please wait to use the spin wait icon that is used throughout the rest of the site.

src/encoded/static/components/login.js
```diff
@@ -114,7 +105,7 @@ loginToServer: function(data) {
   handleSubmit: function(e){
     e.preventDefault();
     var username = this.state.username.trim();
-		var password = this.state.password.trim();
+	var password = this.state.password.trim();
```
j1z0: tiny nitpick...
indentation seems a bit out of place here..

src/encoded/static/components/mixins.js
```diff
@@ -572,9 +600,11 @@ module.exports.HistoryAndTriggers = {
...
                 window.history.pushState(null, '', response_url + fragment);
             }
-            this.setProps({
-                href: response_url + fragment
-            });
+            // store.dispatch({
+            //     type: 'href',
+            //     value: response_url + fragment
+            // });
+            dispatch_dict['href'] = response_url + fragment;
```
j1z0: ok so we discussed this a bit in the meeting today, I guess lets go with codacy's convention and use . instead of []

src/encoded/static/components/navigation.js
```diff
@@ -18,6 +18,12 @@ var Navigation = module.exports = React.createClass({
 
     getInitialState: function() {
         return {
+            testWarning: false
+        };
+    },
+
+    getInitialState: function() {
+        return {
             testWarning: !productionHost[url.parse(this.context.location_href).hostname]
```
j1z0: globals.productionHost should be updated to data.4dnucleome.org

src/encoded/static/components/app.js
j1z0: in fact this should probably read from globals.productionHost.

PR #240, opened 2016-10-12
FF-170 #comment Adding & enabling icon webfonts...

alexkb0009:
+ Replacing '+'/'-' in browse view with up/down angles, related CSS.


alexkb0009:
Requires ` bin/compass compile ` to be run to re-compile style.css from layout.scss (minor edits re: experiment-set button/icon) (unsure if automated).

PR #244, opened 2016-10-16
Minor edits to about page & statics.js formatting

alexkb0009:
(Probably not high priority)
Bold-ing team member names & replacing quotes with backticks to allow multiline text in src/encoded/data/statics.js file (ES6 feature) (y/n?)


alexkb0009:
Ignored 24 'undef' undefined variables issues in Codacy in statics-test.js

PR #247, opened 2016-10-17
Exporting Experiment (Sub-)Table into own file & component

alexkb0009:
@j1z0 @carlvitzthum 
Ready for master (I think) (to pull in updated version) & works well on Browse page (same functionality/appearance as before); may edit `experiments-table.js` further to integrate with exp-set page (likely on branch FF-314 instd. of this one).

Moved out code related to experiments/files' data fitting into table into new component (and file) `ExperimentsTable`. Tiny amount of refactoring for comprehensibility. The `experiments-table.js` module now also contains a function `getFileDetailContainer` which houses Carl's code related to parsing experiment/file data for table that may be imported to other modules/component. `ExperimentsTable` performs same actions as ExperimentSet/ExperimentSetRow previously did in terms of setting state of `selectedFiles`, but also allows to supply a `parentController` react element (via `parentController={this}` where `this` is the parent component's reference to itself so that the parentController may be used to manage state (state.checked, state.selectedFiles) for integration with Browse view & checkboxes and similar functionality later on Exp-Set page.

### DIFF
--- a/src/encoded/static/components/browse.js
+++ b/src/encoded/static/components/browse.js
@@ -15,6 +15,8 @@ var ButtonToolbar = require('react-bootstrap').ButtonToolbar;
 var DropdownButton = require('react-bootstrap').DropdownButton;
 var MenuItem = require('react-bootstrap').MenuItem;
 var store = require('../store');
+var ExperimentsTable = require('./experiments-table').ExperimentsTable;
+var getFileDetailContainer = require('./experiments-table').getFileDetailContainer;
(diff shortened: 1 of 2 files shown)


alexkb0009:
@carlvitzthum Is there a way (e.g. script in bin maybe?) to generate extra data for experiments/sets in DB? My local has 1 exp-set + 4 experiments visible so haven't tested 100% (e.g. edits in commit 19854ca).

PR #310, opened 2016-12-08
inserts for workflow stuff

j1z0:


SooLee:
How about moving these inserts to a different directory, so that they won't get duplicated? (I had already posted them to dev server using fdnDCIC.py (with different uuid))

j1z0:
well they only got duplicated cause you posted them manually, in the future that shouldn’t happen and we always want to make sure they are in the system, so the inserts are in the correct place right now, you can change the status to DELETED for the items you already posted.

SooLee:
My fault not telling you clearly.. (and placing the json in Submit4dn instead). What if I change the uuids to match the uuids in the system?

SooLee:
By the way, as we run workflows, workflow_run objects will be posted manually through lambda using fdnDCIC (the inserts won't be in tests/data/inserts on fourfront.) I thought in the same way about workflow and workflow-related files.

PR #319, opened 2016-12-16
Workflow5

SooLee:
Some example fastq inserts and modified workflow_run link in file_processed to WorkflowRun.


SooLee:
These changes are necessary for the current tibanna lambda tests to work.

PR #327, opened 2017-01-18
search and selective embedding and all that jazz (there is associated snovault branch calle new_indexing_embedding)

j1z0:


src/encoded/static/components/collection.js
```diff
@@ -169,21 +169,18 @@ var lookup_column = function (result, column) {
...
 
         guessColumns: function (props) {
-            var column_list = props.columns || props.context.columns;
             var columns = [];
-            if (!column_list || Object.keys(column_list).length === 0) {
-                for (var key in props.context['@graph'][0]) {
-                    if (key.slice(0, 1) != '@' && key.search(/(uuid|_no|accession)/) == -1) {
+            // Get columns right from results. Selective embedding defines cols used
+            for (var key in props.context['@graph'][0]) {
+                if (key.slice(0, 1) != '@' && key.search(/(uuid|_no|accession)/) == -1) {
```
j1z0: probably a comment about leaving out special fields here would be helpful for the next guy that has to debug this stuff..

src/encoded/static/components/collection.js
```diff
@@ -169,21 +169,18 @@ var lookup_column = function (result, column) {
...
                     }
                 }
-                columns.sort();
-                columns.unshift('@id');
-            } else {
-                for(var column in column_list) {
-                    columns.push(column);
-                }
             }
+            columns.sort();
```
j1z0: just curious is it a requirement that columns are sorted?

src/encoded/static/components/search.js
```diff
@@ -509,22 +508,16 @@ var ResultTable = search.ResultTable = React.createClass({
         }
         // Check to see if we are searching among multiple data types
         // True if only facet is of field "type" when ignoring audits
-        var facet_types = [];
         for (var i = 0; i < facets.length; i++){
-            if (facets[i]['field']){
-                if (!(facets[i]['field'].includes("audit"))){
-                    facet_types.push(facets[i]['field'])
+            if (facets[i]['field'] && facets[i]['field'] == 'type'){
```
j1z0: it is worth breaking this out into a function called get_facet_filter perhaps just for readability?

src/encoded/tests/test_search.py
```diff
@@ -5,54 +5,13 @@
 
 
 def test_search_view(workbook, testapp):
```
j1z0: let's add a test to ensure /search automatically as ?type=Item if nothing specified

PR #412, opened 2017-03-27
FF-660

carlvitzthum:
Includes:
- FF-659: different create and clone forms (clone only applies to experiment and biosample)
- front end md5 calculation
- submission detail changes on edit
- other minor front end file upload changes


src/encoded/static/components/util/file-utility.js
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+Adapted from http://stackoverflow.com/questions/39112096
+Takes a file object and optional callback progress function
+*/
+var getLargeMD5 = module.exports.getLargeMD5 = function(file, cbProgress) {
+    return new Promise((resolve, reject) => {
+        var md5 = CryptoJS.algo.MD5.create();
+        readChunked(file, (chunk, offs, total) => {
+            md5.update(CryptoJS.enc.Latin1.parse(chunk));
```
j1z0: I'm fairly certian you want utf-8 encoding and not latin1

PR #414, opened 2017-04-01
FF 645 :: Browse Table Styling, Analytics, Chart Cursor Popover

alexkb0009:


alexkb0009:
Any remaining Codacy issues should be fine (destructuring expression, 'require' not defined, 'ga' not defined, & similar) though might be worth a quick double-check.

PR #486, opened 2017-06-15
FF-775 :: User-selection of Columns, Table Expansion, Selected Files

alexkb0009:
Should be all set -- want to get into master and have people test over time for especially performance and UI responsiveness of table.


carlvitzthum:
Looks awesome! There are outstanding issues with the search page--namely, facets not working due to double encoding of filter terms--but these have been addressed in another branch.

PR #503, opened 2017-06-23
FF-828

carlvitzthum:
FF counterpart to snovault FF-828 branch, which adds testing for default embedding and manual embedding. In fourfront_utils, default embeds are added to automatically get all fields within subobjects of a schema.


j1z0:
@carlvitzthum this can no longer be merged because of what we talked about yesterday as having fourfront point to snovault branch master from now on... I think if you just update that then it's all good to merge

PR #517, opened 2017-07-11
Update gen ont

aschroed:
Hey can you have a look at this (provided it passes) and merge if it looks OK.  I just modified a bit how terms are compared to see if things have changed which should result in updates being smaller in general.

### DIFF
--- a/src/encoded/commands/generate_ontology.py
+++ b/src/encoded/commands/generate_ontology.py
@@ -373,9 +373,12 @@ def get_slim_terms(connection):
     slim_terms = []
     for cat in slim_categories:
         terms = get_FDN(None, connection, None, search_suffix + cat)
-        if terms.get('notification') is None:
+        try:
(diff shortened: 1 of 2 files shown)


src/encoded/commands/generate_ontology.py
```diff
@@ -456,6 +456,17 @@ def verify_and_update_ontology(terms, ontologies):
...
 
 
+def _get_t_id(val):
+    # val can be: uuid string, dict with link_id, dict with uuid if fully embedded
+    try:
+        linkid = val.get('link_id')
+        if linkid is None:
+            linkid = val.get('term_id')
+        return linkid
+    except:
```
j1z0: except AttributeError maybe... would keep the static analysis gods happy

aschroed: Done.

PR #520, opened 2017-07-13
Item Detail List Updates : Tables for Embedded Objects (v0.33) + Indentation

alexkb0009:
Precursor to remainder of branch FF-818, which is in PR https://github.com/4dn-dcic/fourfront/pull/522 . Can review both PRs by looking at 522, or do in segments (this PR, then 522).

For 'Tables for Embedded Objects' -- still need to handle lists of linkTo objects, which should use their schema column properties, if available, or revert to using expandable list or simple link for it.
Also can use CSS-type iteration.

Organizational Edits: Renaming & moving of `/static/components/` `browse.js` and `search.js` to `/static/components/browse/` `BrowseView.js` and `SearchView.js`, is complete.


alexkb0009:
-- DO NOT MERGE --

Fails a couple of tests, fixed in FF-818 branch which combines this branch and other edits into another PR ( https://github.com/4dn-dcic/fourfront/pull/522 ). Checkout that PR/branch instead.

PR #573, opened 2017-08-29
Tracing Workflow Runs of a Processed File, Draft 1

alexkb0009:
Live at e.g. http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/files-processed/4DNFIYNMI8AD/

Tracing endpoint is e.g. http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/trace_workflow_run_steps/e28632be-f968-4a2d-a28e-490b5493bdc2/ for same file. Currently requires a UUID, should support more unique IDs (accession, etc.) in future.

Can measure performance using url param 'track_performance' e.g. http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/trace_workflow_run_steps/e28632be-f968-4a2d-a28e-490b5493bdc2/?track_performance=True . Add param 'all_runs=True' to get all steps (no grouping of runs by Workflow).


src/encoded/types/workflow.py
```diff
@@ -135,6 +140,367 @@
...
+            in_file['@type'] = input_file_model_obj.get('@type')
+            in_file['file_type'] = input_file_model_obj.get('file_type')
+            in_file['filename'] = input_file_model_obj.get('filename')
+            # Get @ids from ES source.
+            output_of_workflow_runs = input_file_model_obj.get('workflow_run_outputs', [])
+            if len(output_of_workflow_runs) == 0:
+                continue
+            # There should only ever be one 'workflow_run_outputs' at most, or versions of same one (grab most recent).
+            last_workflow_run_output_of = output_of_workflow_runs[len(output_of_workflow_runs) - 1]
+            workflow_run_uuid = get_unique_key_from_at_id(last_workflow_run_output_of)
```
j1z0: seems like these four lines could all be part of get_model_by_uuid .. maybe just call it get_model_by_atid?

src/encoded/types/workflow.py
```diff
@@ -135,6 +140,367 @@
...
+                        'workflow_steps'    : workflow_model_obj.get('workflow_steps'),
+                        'uuid'              : workflow_uuid,
+                        'date_created'      : workflow_model_obj.get('date_created'),
+                        'link_id'           : workflow_model_obj.get('link_id'),
+                        'workflow_type'     : workflow_model_obj.get('workflow_type'),
+                        'cwl_pointer'       : workflow_model_obj.get('cwl_pointer')
+                    }
+
+
+        # Add Output Files, 1-level deep max (maybe change in future)
```
j1z0: this is basically same code as for input files... probably refactor into resuable function and then it should be easier to write some tests

src/encoded/types/workflow.py
```diff
@@ -135,6 +140,367 @@
...
+    "track_performance" : False
+}
+
+
+def trace_workflows(original_file_item_uuid, request, file_item_input_of_workflow_run_uuids, file_item_output_of_workflow_run_uuids, options=None):
+
+    if options is None:
+        options = DEFAULT_TRACING_OPTIONS
+
+    if options.get('track_performance'):
```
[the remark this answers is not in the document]
alexkb0009: Cool! Thx, was wondering w/ Carl how to potentially do that.

src/encoded/loadxl.py
```diff
@@ -803,6 +803,24 @@ def load_prod_data(app, access_key_loc=None):
...
+        'REMOTE_USER': 'TEST',
+    }
+    testapp = TestApp(app, environ)
+
+    from pkg_resources import resource_filename
+    inserts = resource_filename('encoded', 'tests/data/jin_inserts/')
+    docsdir = []
+    load_all(testapp, inserts, docsdir)
+    keys = generate_access_key(testapp, access_key_loc,
+                               server="https://testportal.4dnucleome.org")
```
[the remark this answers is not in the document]
alexkb0009: Will change to data.4dnucleome (?) maybe in future could grab from environment variable (?). IDK.

src/encoded/types/workflow.py
```diff
@@ -135,6 +140,367 @@
...
+        filtered_out_tuples = []
+        tuples_by_workflow = {}
+        for workflow_run_uuid, in_file_uuid, workflow_run_model_obj, in_file in workflow_run_tuples:
+            workflow_atid = workflow_run_model_obj.get('workflow')
+            if not tuples_by_workflow.get(workflow_atid):
+                tuples_by_workflow[workflow_atid] = []
+            tuples_by_workflow[workflow_atid].append((workflow_run_uuid, in_file_uuid, workflow_run_model_obj, in_file))
+
+        def get_date_created_from_tuple(wfr_tuple):
+            workflow_run_uuid, in_file_uuid, workflow_run_model_obj, in_file = wfr_tuple
```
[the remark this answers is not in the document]
alexkb0009: I think the same-name`filter_workflow_runs` variables aren't overwritten since the `for` block ends before the func definition and hopefully are de-referenced (or at least that was my intent lol). But ya either way don't have any actual need to unpack the tuple (other than to access the value). Will try to improve.

src/encoded/types/workflow.py
alexkb0009: I need to rename that func lol. Reasoning I kept it explicit is that sometimes can grab a UUID from atid, sometimes can only grab accesssion. When use accession, need to use `get_by_unique_key(accession, 'accession')` rather `than get_by_uuid`, and get_by_unique_key returns a different type of Item (Snovault Item I believe) than get_by_uuid, which returns a model. If don't make clear distinction (uuid vs accession), then might become more confusing to determine if are dealing with a SnowyItem (via accession) or ORM model (via UUID). Eventually I'm thinking could just make it request.embed() (the `get_by_atid`) once/if performance improves and delete a lot of lines of code?

alexkb0009: Oops nevermind.... I thought GH comments were about the lines above the comment not below. Yeah definitely, will do.

src/encoded/types/workflow.py
alexkb0009: Will do. Likely will wait until get to writing trace_future function. Reasoning is that when tracing history, we also trace input files (see step['inputs']['source'] -- calls recursive tracing function -- vs step['outputs']['target'] -- where just have list with 1 item until maybe a subsequently traced input files 'claims/matches' it and adds to it in lines 460-472). I haven't written `generate_targets_for_output` yet (or converted `generate_sources_for_input` to work for both).

src/encoded/loadxl.py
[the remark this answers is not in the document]
alexkb0009: Actually, mastertest is in there for load_test_data, so will re-use that for load_jin_data URL.

src/encoded/types/workflow.py
alexkb0009: Actually... I do seem to be getting model from get_by_unique_key, well then, this simplifies things.... :-D

PR #617, opened 2017-09-20
Small Edits to BiosampleView; Full Screen Workflow View (Draft); Workflow Schema Updates

alexkb0009:
Use overview section in place of table if only 1 biosource.
May be improved somewhat further.
ToDos Remaining:
- List of Experiments or Sets used in. Will likely need rev-link.
- List of related biosamples. May wait until we have example.

## Some re-indexing will occur upon deployment as 'analysis_steps' schema has been somewhat updated, as well as some embeds.

**Adding embed for workflow_run.workflow.workflow_steps.step.software_used**


j1z0:
this is now outdated... closing

PR #622, opened 2017-10-12
Front-end Cleanup & Polish for Workflow Graphs & Related Components

alexkb0009:


src/encoded/static/components/viz/Workflow/Graph.js
```diff
@@ -110,25 +112,30 @@ export default class Graph extends React.Component {
             });
             return nodes;
         },
+        /**
+         * This function, along with 'nodesInColumnPostSortFxn' & 'nodesPreSortFxn', is _PROTOTYPICAL_ - a R&D work in progress.
+         * It's quick, dirty, hacky. And it will remain this way until we figure out what we want from it, or if we want it at all, and then we'll go back and make it DRY and pretty.
+         */
         'nodesInColumnSortFxn' : function(node1, node2){
```
alexkb0009: This part _should_ be cleaned up at some point, but it won't for a while because is just experimental row positioning at this point. If find approach we like, will then cement/clean/refactor.

src/encoded/static/components/viz/Workflow/Node.js
```diff
@@ -11,8 +12,7 @@ import { Fade } from 'react-bootstrap';
 export class DefaultNodeElement extends React.Component {
```
alexkb0009: This Component/Class isn't used by us but only exists for if we publish to NPM. Will likely need a tiny bit of clean-up; along with saving SCSS into NPM lib. For later / low priority.

src/encoded/static/components/viz/Workflow/Node.js
```diff
@@ -230,10 +234,6 @@ export default class Node extends React.Component {
         */
     }
 
-    static propTypes = {
-        'title' : PropTypes.func.isRequired
```
alexkb0009: We no longer need this, because output of `renderNodeElement(node, graphNodeProps)` (aka our WorkflowNodeElement) should handle all of that (titles on node).

src/encoded/static/components/viz/Workflow/parsing-functions.js
```diff
@@ -4,34 +4,165 @@ import _ from 'underscore';
...
+    /***************
+     ** Functions **
+     ***************/
+
+     /**
+      * @param {StepIOArgument} stepIOArg - Input or output argument of a step.
+      * @param {boolean} [readOnly=true] - @see preventDuplicateNodeID()
+      * @returns {string} Unique ID for node.
+      */
+    function ioNodeID(stepIOArg, readOnly = true){
```
alexkb0009: Step (& step nodes) must be unique in their 'name' property (aka steps.name, stepNode.name).
IO arguments *do not* need to be unique on their name _except within a step_ (CWL struct anyway I think) and will auto-generate an additional ID, because there might be multiple steps with same argument name, in addition to multiple files for one argument. Autogenerated ID is just `steps['inputs'||'outputs'].name[ + '~' +
 <duplicate_increment>]`, it might be worthwhile to change this to `steps.name + steps['inputs'||'outputs'].name[ + '~' +
 <duplicate_increment>]`. Doesn't affect graphing, just for semantics/data-completeness/correctness.

src/encoded/static/components/viz/Workflow/parsing-functions.js
```diff
@@ -46,10 +177,10 @@ export function parseAnalysisSteps(analysis_steps, parsingMethod = 'output'){
     }
 
     function compareTwoFilesByUUID(file1, file2){
```
alexkb0009: We currently still allow `steps.inputs.run_data.file` to be in form of either a UUID or an embedded object.
We could simplify to only handle embedded object to make this front-end code _slightly_ simpler, after transforming 'file' to a linkTo and ensuring its consistent embeddedness, however there might be a benefit to this type-ambiguousness if publishing to NPM, as others' data/references might differ.

src/encoded/static/components/viz/Workflow/parsing-functions.js
```diff
@@ -219,7 +344,7 @@ export function parseAnalysisSteps(analysis_steps, parsingMethod = 'output'){
 
             var groupKeys = _.keys(groups);
             var filesNotInGroups = [];
-            var filesByGroup = _.reduce(stepInput.run_data.file, function(m, file, idx){
+            var filesByGroup = _.reduce(stepIOArgument.run_data.file, function(m, file, idx){
                 var incl = false;
```
alexkb0009: This entire section of code (grouping of files) works but is unwieldy; will likely be re-done completely at a future time but kept in non-refactored/cleaned-state for now.

src/encoded/static/components/item-pages/components/WorkflowDetailPane/FileDetailBody.js
```diff
@@ -0,0 +1,303 @@
...
+import { Fade } from 'react-bootstrap';
+import { console, object, layout, ajax, fileUtil, expFxn } from './../../../util';
+import { getTitleStringFromContext } from './../../item';
+import { FlexibleDescriptionBox } from './../FlexibleDescriptionBox';
+import { SimpleFilesTable } from './../SimpleFilesTable';
+import { ItemDetailList } from './../ItemDetailList';
+import { ExperimentSetTablesLoaded } from './../ExperimentSetTables';
+import { ViewMetricButton, MetricsView } from './FileDetailBodyMetricsView';
+
+
```
alexkb0009: DetailPanes are not super important at moment and will likely change in future.

Currently works somewhat like decision tree of components.
<WorkflowDetailPane/> renders one of [FileDetailBody, StepDetailBody, WFRDetailBody, ParameterDetailBody]

Each of those reads data off of selectedNode.meta for everything it needs. Will perform AJAX of @id to get more info, etc. If is QC metric, FileDetailBody will render some additional rows for QC stuff, etc.

src/encoded/static/components/viz/Workflow/Graph.js
```diff
@@ -254,12 +283,35 @@ export default class Graph extends React.Component {
                         nodesInColumn.splice(afterThisIdx + 1, 0, gN);
                     }
                 });
-                // Update own indices (not used for anything currently xcept debugging)
-                _.forEach(nodesInColumn, function(n, index){
-                    n.indexInColumn = index;
-                });
             }
             
+            if (_.every(nodesInColumn, function(n){ return n.type === 'step'; })){
```
alexkb0009: This part will very likely change in future.

Currently if any step node(s) in column of step nodes has more inputs than node w/ lowest # of inputs in same column: gathers such nodes and injects them into middle of column.

Doesn't do anything smarter than that at the moment. One possibility for the future could be to try to calculate subsequent edge intersections arising from placement at each index and then select placement to minimize it.

PR #631, opened 2017-10-25
FF-985

carlvitzthum:
Foursight changes on health page:
* localized timestamps (no longer UTC)
* foursight controls are admin-only, but the rest is viewable to everyone
* made onClick functions more robust
* error messages are now displayed in JSON tree form rather than logged to the console

Submissions page:
* added ExperimentDamid
* fixed issue with cancel selection button when on "select existing" page


src/encoded/static/components/item-pages/HealthView.js
```diff
@@ -90,60 +90,74 @@ export class HealthView extends React.Component {
...
 
 
-class AdminPanel extends React.Component {
+class FoursightPanel extends React.Component {
 
     constructor(props){
         super(props);
-        this.loadFoursight = _.throttle(this.loadFoursight, 500);
-        this.runFoursight = _.throttle(this.runFoursight, 500);
+        this.foursight_server_by_stage = {
```
j1z0: why did this get moved back here from the ini files?

PR #632, opened 2017-10-26
FF-895

carlvitzthum:
Added aggregation on missing/null values. The placeholder is 'No value'. Faceting on this is supported, as well as setting NOT (my_field!=No+value)

Test added

A good test case: faceting on Experiment HiC by digestion enzyme or publication.


aschroed:
ajs [3:37 PM] 
Looks good.  Things I notice:
1. can only be accessed via URL searches at the moment, right?
2. if you enter a field that is not a field of that item type all items of that type are returned
eg. http://localhost:8000/search/?cell_culture_details.culture_start_date=No+value&type=Experiment
3. similarly if you enter a field of an item that is a linkTo field (without using a specific (embedded?) field) all the items are returned
eg. http://localhost:8000/search/?cell_culture_details=No+value&type=Biosample

I’m not sure that these are problems as such and if I do well behaved searches it works as expected.  Just thought I’d point them out.

PR #638, opened 2017-11-09
ImagingV2

KorayKirli:


aschroed:
Looks good.  I made a couple of typo tweaks you should confirm are ok.

2 things:
1. I didn't see that any of the file_microscopy inserts that had links to microscopy_settings inserts?  Maybe want to add at least one?

2. You know I've gotta ask this - will there never be more than 3 channels in microscope settings?  Listing specific fields 'redundantly' for each channel is like fingernails on a blackboard to me but I can see how if there will never be more than the 3 then it makes things more straightforward rather than having embedded sub objects or specific linkTos to a channel schema.

You can go ahead and merge once you think it's good to go.

PR #673, opened 2018-01-02
Frontend Code Cleaning & Minor Iteration

alexkb0009:
### Changes

- Removing majority of ‘high-level’ instances/references/etc of ‘expSetFilters’ and functions/code related to it (most of work).
- Replacing storage of `expSetFilters` structured data in Redux Store with a function `contextFiltersToExpSetFilters(contextFilters = ReduxStoreState.context.filters)` to generate same structure — primarily for backward compatibility with existing visual components & functions which utilize `expSetFilters` structure (e.g. `Filters.changeFilter(field, term, expSetFilters)`). **N.B.** Originally idea was to have ‘hrefToExpSetFilters’ function (instead of ‘contextFiltersToExpSetFilters’) - this is deprecated because it’s simpler and ‘more correct’ to work with context.filters returned as part of our search.py (e.g. no need to dupe/hardcode ‘exclusions’ to URL query params on front-end when handle ‘href’).
- Gravatar for user menu item (top right of navbar) - https://gyazo.com/f35b0dffcf73cb368bb77288bdfa666b - and in /search/ table for User item title.
- https://gyazo.com/cbe6d0500a23f68d96f727ed70e4b0f6 should no longer happen.
   - On subscription to Redux href changes, when detects changed filters, also clears out any pop-overs. Will likely be reimplemented somewhat if/after move chart data aggregation to back-end. Probably by keeping subscription to Redux store Href but moving somewhere else (app… maybe).
- Added 'excludedColumns' optional field to Item schemas which works alongside 'columns' field to hide default-visible columns such as `lab.display_title`.
- RawFilesStackedTable will show 'no view permission' for appropriate Items on Item views instead of non-functional link (`<a href="#">...<.a>` was previous fallback, is now `<em>{ item.error }</em>`).
- Removed FacetList component from ExpSetView (as it can no longer filter client-side) and client-side filtering code. Replaced with 'overview heading' -> https://gyazo.com/030e49dd378304cb1229384cbd07d5dd
- Changed search.py `get_all_results` function into 2 separate ones to take advantage of generators. Why? Mostly preparing-for/exploring ‘back-end aggregation of chart data’ out of curiosity to see how much work it’d involve. `es_result['hits']['hits']` as returned is now a generator and not a list (until is run through `list(graph)` or whatever).

---

Off-Topic-Ish /  **thinking out load for future** (further plans for this should be discussed if going to have aggregation on back-end). _Previous_ (deprecated) thoughts:
> In search.py on browser requests (not load-as-you-scroll, nor JSON, nor subrequests), perform copy of /search/ request but w/ *limit = all*, then aggregate into _various_ structures which needed on front-end. Such structures vary depending on chart and it might look like the following snippet at the _simplest_ … 

... Once I realized the number of combinations of aggregations that would need to compute to meet all x/y axis and subdivision field options… starting to think a better approach might be a custom AJAX endpoint(s) that takes URL params for e.g. ‘x’ field, ‘subdivision’ field, etc. Anyways not too important ATM there’s other ToDos before this comes close to horizon.

```
context : {
    ‘@graph’ : […],
    …
    ‘aggregations’ : {
        // For QuickInfoBar
        ‘counts’ : {
            ‘filtered_files’ : 100,
            ‘filtered_experiments’ : 20,
            ‘filtered_experiment_sets’ : 5,
            ‘total_files’ : 200,
            ‘total_experiments’ : 45,
            ’total_experiment_sets’ : 10
        },
        // For BarPlotChart
        ’counts_by_field_term_experiment_sets’ : {
            ‘award.project’ : {
                // X axis counts/grouping * all possibilities (fields available for ‘X axis’ dropdown)
                ‘external’ : {
                    // Subdivision counts… * all possibilities (fields available for ‘Group By’ dropdown)
                    ‘experiments_in_set.biosample.biosource.individual.organism.name’ : {
                        ‘mouse’ : 1,
                        ‘human’ : 1
                    },
                    ‘experiments_in_set.experiment_type’ : {
                        ‘In-situ HiC’ : 1,
                        ‘Dilution HiC’ : 1,
                        ‘Capture HiC’ : 1
                    },
                    … … …
                },
                ‘4DN’ : {
                     ‘experiments_in_set.biosample.biosource.individual.organism.name’ : {
                        ‘mouse’ : 1,
                        ‘human’ : 2
                    },
                    ‘experiments_in_set.experiment_type’ : {
                        ‘In-situ HiC’ : 0,
                        ‘Dilution HiC’ : 1,
                        ‘Capture HiC’ : 1
                    },
                    … … …
                }
            },
            ‘experiments_in_set.experiment_type’ : {
                ‘In-situ HiC’ : {
                     // Subdivision counts… 
                    ‘experiments_in_set.biosample.biosource.individual.organism.name’ : {
                        ‘mouse’ : 1,
                        ‘human’ : 2
                    },
                    … … …
                 },
                 … … …
            },
            … … …
        },
        ’counts_by_field_term_experiments’ : { ... ... ... ... ... },
        ’counts_by_field_term_files’ : { ... ... ... ... ... }
    }
    …
}

### CODE REVIEW

File: src/encoded/schemas/biosource.json [hunk 1]
```diff
@@ -47,82 +47,81 @@
               "tissue",
               "multicellular organism"
             ]
-          },
```
alexkb0009: Changed: indentation

src/encoded/schemas/experiment_set.json
```diff
@@ -113,7 +113,7 @@
         "experiments_in_set.digestion_enzyme.name": {
             "title": "Enzyme"
         },
-        "experiments_in_set.biosample.modifications.modification_type'": {
```
alexkb0009: Removed typo(?) -- seems to improve facet on local.

src/encoded/schemas/lab.json
```diff
@@ -130,5 +130,20 @@
...
+            "title" : "Institute"
+        },
+        "state" : {
+            "title" : "State"
+        },
+        "country" : {
+            "title" : "Country"
+        }
+    },
+    "excludedColumns" : ["lab.display_title"]
```
alexkb0009: Is used by front-end to remove columns from SearchResultTable.

src/encoded/search.py
```diff
@@ -230,29 +231,28 @@ def get_date_range(request):
     return before, after
 
 
-def get_all_results(search):
```
alexkb0009: Split func into two, ensure returns es_results['hits']['hits'] as generator instead of list if size requested > increment.

src/encoded/static/components/collection.js
```diff
@@ -1,17 +0,0 @@
-'use strict';
```
alexkb0009: File no longer used anywhere.

src/encoded/static/components/index.js
```diff
@@ -34,7 +33,7 @@ require('./item-pages/FileMicroscopyView');
 require('./item-pages/BiosampleView');
 require('./item-pages/BiosourceView');
 require('./item-pages/ProtocolView');
-require('./statuslabel');
+//require('./statuslabel');
```
alexkb0009: Not currently used, keeping file as might re-use later?

PR #681, opened 2018-01-08
'Create/Add' Button Permissions & Collection-Type-Selection

alexkb0009:
'add' action/permission is calculated for an AbstractCollection now instead of Collection, allowing Experiments Collection to answer to permissions requests to see if current user/requestor may add.

Front-end edits to submissions to ensure type is up-to-date in/after initialization.

### DIFF
--- a/src/encoded/static/components/submission/submission-view.js
+++ b/src/encoded/static/components/submission/submission-view.js
@@ -120,7 +120,7 @@ export default class SubmissionView extends React.Component{
      * available.
      */
     componentDidMount(){
-        if(this.props.schemas && Object.keys(this.props.schemas).length > 0){
+        if(this.props.schemas && _.keys(this.props.schemas).length > 0){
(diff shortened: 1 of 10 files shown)


src/encoded/static/components/submission/submission-view.js
```diff
@@ -534,8 +533,8 @@ export default class SubmissionView extends React.Component{
             newHierarchy = modifyHierarchy(hierarchy, keyIdx, parentKeyIdx);
             validCopy[keyIdx] = 1; // new object has no incomplete children yet
             validCopy[parentKeyIdx] = 0; // parent is now not ready for validation
-            typesCopy[keyIdx] = type;
         }
+        typesCopy[keyIdx] = type;
```
alexkb0009: Important-est front-end edit. Everything else is minor and unimportant (extraState can be reverted).

PR #682, opened 2018-01-09
More inserts

SooLee:


alexkb0009:
There is some front-end clean-up that could be done around the 'Item Page Components' relating to Workflows (not actual parsing/drawing functions), which am holding off on until merge to master as cleanup of this would conflict with https://github.com/4dn-dcic/fourfront/pull/685

PR #730, opened 2018-03-08
FF-1068

carlvitzthum:
# Adding last_modified

last_mofieid is a schema mixin with default value of [] (mixin name is `modified`). It is an object containing `date_modified` and `modified_by`, so we have a record of the time and user that lost modified an item.

Every item that uses the submitted mixin also uses the modified mixin.


j1z0:
@carlvitzthum I almost flipped when I saw 77 files changed then I realized it was just schemas :P.  Anyways its looks pretty good, but a couple of comments.  The Codacy issues I think are fine, you can just login to Codacy and set it to ignore the assert issue altogether... as for coverage though, can you do anything about the decrease in coverage?

PR #735, opened 2018-03-12
Testing Multi-level Sorting

alexkb0009:


src/encoded/search.py
```diff
@@ -103,6 +104,11 @@ def search(context, request, search_type=None, return_generator=False, forced_ty
     ### Adding facets to the query
     search = set_facets(search, facets, query_filters, string_query)
 
+    ### Add preference from session, if available
+    if request.__parent__ is None and not return_generator and size != 'all': # Probably unnecessary, but skip for non-paged, sub-reqs, etc.
```
alexkb0009: AFAIK this 'if' statement is not needed (no tests fail without it, etc), but includes cases where there is no expectation of paging for optimization.

PR #745, opened 2018-03-26
Higlass

j1z0:
should be ready to go


alexkb0009:
It might be worth waiting for HiGlass JS lib to be packaged non-minified before merging re: build times & work on in parallel (UI) w/ master branch?

PR #756, opened 2018-04-04
Schema edits2

sbreiff:
- added more facets and columns to several schemas
- Modification and Treatment facets now show only type (eg, "RNAi" or "Chemical" for treatments, "Stable transfection" or "Crispr" for Modifications)
- Modification and Treatment columns now show full modification/treatment name, or list if there are multiple
- display_title for treatment_chemical and treatment_rnai were edited to make more concise


sbreiff:
Also made changes (incl. for Guttman lab FISH submission):
- added r3d, dv files and file extensions to file_microscopy.json
- added a few exclude_from fields to mixins.json
- added exclude_from 'Submit4dn' to cell_line_tier property in biosample.json

PR #762, opened 2018-04-10
Repliseq bw

SooLee:


j1z0:
merging in master to get tests to pass

PR #771, opened 2018-04-27
(Approaching Readiness) Static Pages Hierarchy & DropDown

alexkb0009:
Lots of new features.

Remaining ToDos include updating [Static Page Documentation](https://github.com/4dn-dcic/fourfront/blob/master/docs/static-pages.md) & Tangential Submission View things.

Working Draft of:

## Best Practices

- DO split Pages into multiple StaticSections with proper title for each, if possible, rather than having Page that has just one big long Markdown section/file.
   - This will allow each section to be re-used in other places & apply permissions to each section.
   - If there is only one or less sections with a title (e.g. could have multiple sections all with no titles or just one big long section), then the `##` (h2) headers get promoted as if they were Section headers in TableOfContents. However, styling within the page itself will remain as Markdown h2 header (not section header). H1 (`#`) headers are reserved for Page titles and are not currently supported within (our parsing of) Markdown.
   - If have *2+* static sections with titles, all sections and their titles — even if nonexistent — will be displayed in TableOfContents. If there’s a section for which title doesn’t exist, will default to “Table of Content Entry”, e.g. [this example](https://gyazo.com/e3f718273d5d46e9ffcab884b9a36fa3) containing 4 sections, with titles for just two of them.
     - **Update**: Default, if no title on Section, is now (JS equivalent of) `" ".join([ word.capitalize() for word in section.link.split("-") ])` where section.link is last bit of StaticSection name (e.g. “lorem-ipsum-1” => “Lorem Ipsum 1”)
- If are going to edit Pages/Sections through Fourfront UI (rather than using a Markdown/text editor & then adding to inserts) — then is a good idea to keep inserts up-to-date in order to make local development + testing simpler as well as provide an extra source of backups.
  - Our primary mission isn’t to maintain/support a custom content management system so having a concrete outside-of-db representation of static pages I think is desireable.
  - There is now a command called `bin/export-data` which can be used to export Page and StaticSection inserts into JSON files. Examples:
    ```bash
    bin/export-data "https://data.4dnucleome.org/search/?type=Page&limit=all" -u ACCESS_KEY_ID -p ACCESS_KEY_SECRET > new_page_inserts_file.json
    bin/export-data "https://data.4dnucleome.org/search/?type=StaticSection&limit=all" -u ACCESS_KEY_ID -p ACCESS_KEY_SECRET > new_static_section_inserts_file.json
    ```


src/encoded/commands/export_data.py
```diff
@@ -0,0 +1,102 @@
...
+
+This has not been tested, but theoretically should be able to pipe inserts from export_data to import_data (e.g. transfer data to different server) for __small__ amounts of data (due to lack of streaming).
+'''
+
+
+def run(search_url, username='', password=''):
+
+    url = urlparse(search_url)
+
+    if url.scheme not in ('http', 'https') or '/search/?type=' not in search_url:
```
carlvitzthum: Might it be better to check for "/search/" and "type=" in url separately? Query param order is arbitrary

PR #774, opened 2018-04-27
Procfile edits

sbreiff:
- file_classification now generally excluded from submit4dn for most file schemas, default values assigned
- csv added to file formats in file_reference.json, mainly to accomodate submitter-specific file formats in microscopy processed files
- reference_fasta and oligo_fasta changed to compressed_fasta and fasta so that fasta files of various purposes can be either fasta.gz or fasta


aschroed:
Looks fine to me.

PR #777, opened 2018-05-04
Barplot Endpoint GET -> POST ("CUT&RUN" param value fix)

alexkb0009:


src/encoded/static/components/viz/chart-data-controller.js
```diff
@@ -31,7 +31,7 @@ var refs = {
     store       : null,
     href        : null, // Cached from redux store updates // TODO: MAYBE REMOVE HREF WHEN SWITCH SEARCH FROM /BROWSE/
     contextFilters : [],
-    baseSearchPath : null,
+    baseSearchPath : '/bar_plot_aggregations',
```
alexkb0009: This actually doesn't ever change; we could move/remove/hardcode-everywhere to simplify code instead of storing in 'refs' dict.

PR #803, opened 2018-06-05
File edits for barcode files (and file_calibration)

sbreiff:
- added reference_files field to experiment.json schema, so that (for example) a barcode file can be associated with the experiment
- added new options in file_format, file_type, and file_format_file_extension in file_reference.json schema to allow for barcode files

unrelated to above: removed Microscopy as default file_classification of file_calibration items - would like to patch current items to be 'ancillary file' and remove Microscopy from file_classification enum eventually


src/encoded/schemas/experiment.json
```diff
@@ -198,6 +198,18 @@
                 }
             }
         },
+        "reference_files": {
+            "title": "Reference Files",
+            "description": "Reference files associated with this experiment (ex. file of sequencing barcodes).",
+            "type": "array",
+            "lookup": 415,
+            "items": {
+                "title": "Processed File",
```
aschroed: Probably want to call this Reference file.

PR #819, opened 2018-07-02
Replaced

KorayKirli:


src/encoded/tests/test_permissions.py
```diff
@@ -515,11 +514,56 @@ def test_submitter_can_view_ownitem(ind_human_item, submitter_testapp, wrangler_
...
+    my_accession = '/' + res.json['@graph'][0]['accession']
+    submitter_testapp.get(my_accession, status=404)
+
+
+def test_submitter_can_view_ownitem_replaced_using_uuid(ind_human_item, submitter_testapp, wrangler_testapp):
     res = submitter_testapp.post_json('/individual_human', ind_human_item, status=201)
     wrangler_testapp.patch_json(res.json['@graph'][0]['@id'], {"status": "replaced"}, status=200)
-    submitter_testapp.get(res.json['@graph'][0]['@id'], status=404)
+    my_uuid = '/' + res.json['@graph'][0]['uuid']
+    submitter_testapp.get(my_uuid, status=301)
```
aschroed: The 301 is a 'moved permanently' response ah I see so it redirects to the new identifier/url.  Maybe it's worth testing that the url is indeed referring to the same item?

Maybe this is the best we can do in unit testing for 301 redirects but when you get a 301 you really aren't getting the item.

KorayKirli:
Reformatted the tests to trace the redirect and make sure that we get 200.
It turns out the redirect for uuid is to /object/uuid.

PR #838, opened 2018-07-23
Tsa tweaks

aschroed:


src/encoded/schemas/experiment_seq.json
```diff
@@ -53,7 +55,8 @@
             "GAM",
             "CUT&RUN",
             "TrAC-loop",
-            "TSA-seq"
+            "TSA-seq-v1",
```
sbreiff: Remove these two lines from enum

src/encoded/schemas/experiment_tsaseq.json
```diff
@@ -0,0 +1,117 @@
...
+        { "$ref": "mixins.json#/facets_common" }
+    ],
+    "mixinColumns": [
+        { "$ref": "experiment.json#/columns"}
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "experiment_type": {
```
sbreiff: Since this is a TSA-seq-only schema, possibly edit experiment_type to include "default": "TSA-seq" and exclude from Submit4dn and FFedit-create

src/encoded/schemas/file_reference.json
```diff
@@ -42,6 +42,7 @@
                 "compressed_fasta",
                 "fasta",
                 "juicer_format_restriction_site_file",
+                "conc",
```
sbreiff: Is it worth also adding 'spike-in' as a file_type?

PR #847, opened 2018-08-02
Sample type dt

aschroed:
Added biosample_type calculated property on biosample.
Added to experiment and experiment_set embed list
Added 'appropriate' facets.


aschroed:
Added biosample_type calculated property on biosample.
Added to experiment and experiment_set embed list
Added 'appropriate' facets.

src/encoded/schemas/biosample.json
```diff
@@ -137,9 +137,12 @@
         }
     },
     "facets": {
-        "biosource_summary": {
+        "biosample_type": {
```
sbreiff: When looking at this on local, the biosample_type seems a little redundant with biosource_type, even if they're slightly different - might make sense to remove biosource_type from facets, but up to you.

src/encoded/schemas/biosample.json
```diff
@@ -137,9 +137,12 @@
...
             "title": "Sample Type"
         },
+        "biosample_summary": {
+            "title": "Biosample Summary"
+        },
```
sbreiff: I think you meant "biosource_summary" - I didn't see a "biosample_summary" property.

src/encoded/schemas/biosample.json
aschroed: Done.

src/encoded/schemas/biosample.json
```diff
@@ -137,14 +137,14 @@
...
-        "biosource_summary": {
+        "biosample_type": {
             "title": "Sample Type"
         },
         "biosource.cell_line_tier": {
             "title": "4DN Cell Line Tier"
         },
-        "biosource.biosource_type": {
-            "title": "Biosource Type"
+        "biosource.biosource_summary": {
```
sbreiff: biosource_summary is actually a calc prop of biosample, so this should be just "biosource_summary" (like in columns below) instead of "biosource.biosource_summary". Sorry if my previous comment wasn't as clear as it could have been!

src/encoded/schemas/experiment_set.json
```diff
@@ -159,8 +115,11 @@
         "experiments_in_set.biosample.biosource.individual.organism.name": {
             "title": "Organism"
         },
-        "experiments_in_set.biosample.biosource_summary": {
-            "title": "Sample"
+        "experiments_in_set.biosample.biosample_type": {
+            "title": "Sample Type"
+        },
+        "experiments_in_set.biosample.biosample_summary": {
+            "title": "Biosample Summary"
```
sbreiff: Change to "experiments_in_set.biosample.biosource_summary"

PR #848, opened 2018-08-02
Modified badges mixin, and test insert, added in review by lab to bad…

aschroed:
…ge status

I am not sure if we need the badge_purpose bit as part of the subembedded object mixin or not.  Also not sure if 'badge_type' should be added to the badge schema.  Having a field in the mixin gives a bit more flexibility but as you think about types of classes of badges keep the options in mind.

Also added 'in review by lab' status to badge enum.


KorayKirli:
Hey Andy, I see that you took out tests and types from vendor to base, to use same functions for vendor and badge I guess. Just to keep thing tidy, I think we can rename the vendor ones to contain badge also, like "test_types_vendor_and _badge". Not a big deal, so up to you

aschroed:
Koray since they are unit tests and it is feasible though not likely that things could change that might affect one but not the other item type I think I won't combine the tests.  As you said I removed Vendor.py and added the Vendor class back to init to minimize needed imports as they both use the same external functions and one new internal function.

PR #855, opened 2018-08-09
File format item

aschroed:
First steps to converting file_format into an item:

- Added schema
- Added Class
- Added inserts for some file_formats
- Removed file_fasta schema and references to it.


src/encoded/tests/data/inserts/file_format.json
```diff
@@ -0,0 +1,130 @@
...
+        "allowed_extensions": ["bwaIndex.tgz"],
+        "description": "This format is used for bwaIndex files",
+        "uuid": "d13d06cf-218e-4f61-55f0-91f226248b2c",
+        "lab": "4dn-dcic-lab",
+        "award": "1U01CA200059-01"
+    },
+    {
+        "file_format": "compressed_fasta",
+        "standard_file_extension": "fasta.gz",
+        "allowed_extensions": ["fasta.gz", "fsa.gz"],
```
sbreiff: I'd suggest adding ".fa.gz" to this list, since .fa is sometimes used as a fasta file extension.

PR #860, opened 2018-08-15
[BETA VERSION] Stats Aggregation

alexkb0009:
### Current Edits:

It should look like this video: https://gyazo.com/96306465366c071ed52a22976b85daf3

#### Regarding schema updates:
New facets for file size and for date ranges/histograms are introduced **but** they are currently hidden from view in FacetList because there is not yet a mechanism to filter by ranges (numerical or date) outside of text search field. Eventually there will be and we will make these facets visible.

We may want to temporarily xfail test `test_search.py::test_search_facets_and_columns_order` (?): https://travis-ci.org/4dn-dcic/fourfront/jobs/416434749#L2118

### ToDos for V2:

- Legend for chart colors
  - Including better organization of state:
    - Create a "color scale" or "z-axis scale" at a higher level in rendering tree so that multiple charts can utilize the same color for a term if they are e.g. showing the same fields/enum.
    - Probably have it be a re-usable component that can be wrapped around 1 or many different charts.
- Google Analytics, Downloads, etc.
- Better UI for stats page (grid + modal or similar for zoom-in ?)
- Python tests for new aggs & facets (**longer term** - as requirements & aggs may change in short term)


src/encoded/search.py
```diff
@@ -692,153 +691,320 @@ def set_filters(request, search, result, principals, doc_types, before_date=None
...
+        :returns: List of tuples containing (0) ElasticSearch-formatted field name (e.g. `embedded.status`) and (1) list of terms for it.
     """
+
     facets = [
-        ('type', {'title': 'Data Type'})
+        ('type', {'title': 'Data Type'}) # More will be appended to here from schemas & any currently-active filters.
+    ]
+    append_facets = [
+        ('status', {'title': 'Status'}),
+        ('date_created', {'title': 'Date Created', 'hide_from_view' : True, 'aggregation_type' : 'date_histogram' })
```
carlvitzthum: Comment out for now?

src/encoded/schemas/file.json
```diff
@@ -203,13 +203,16 @@
...
         },
         "file_format": {
-            "title" : "File Format"
+            "title" : "File Format",
+            "description" : "Format of the file, i.e. the file extension."
         },
         "experiments.experiment_type" : {
-            "title" : "Experiment Type"
+            "title" : "Experiment Type",
+            "description" : "Type of experiment of which this file is a part of."
```
sbreiff: I know it's just semantics but might look/sound a little better as "Type of experiment to which this file belongs"

src/encoded/schemas/file.json
```diff
@@ -225,6 +228,60 @@
...
+                { "from" : 10240, "to" : 102400 },
+                { "from" : 102400, "to" : 1048576 },
+                { "from" : 1048576, "to" : 10485760 },
+                { "from" : 10485760, "to" : 104857600 },
+                { "from" : 104857600, "to" : 1073741824 },
+                { "from" : 1073741824, "to" : 10737418240 },
+                { "from" : 10737418240, "to" : 107374182400 },
+                { "from" : 107374182400 }
+            ],
+            "disabled" : true,
```
alexkb0009: `disabled` is now used in place of `hide_from_view`, though "hide_from_view" is still supported by front-end if want to include agg for e.g. `format=json`. `disabled` will exclude this facet from being aggregated by ElasticSearch in first place, which should improve performance a smidge.

PR #867, opened 2018-09-05
DO NOT MERGE UNTIL TIBANNA CAN PLAY NICE WITH THIS - File format part 2

aschroed:
Everything appears to work locally and I was able to patch extra files on webdev so am hopeful that we're good to go but this needs a thorough review as it's affecting a lot of moving parts.

Note that as per Soo's request for the time being the file_format_file_extension bits remain in the schemas but they should be redundant and removed once tibanna doesn't use them at all.


KorayKirli:
File type other is not needed anymore, and in the inserts it has the extension "other", which turn upload key to uuid/accession.other . First thought of adding all formats used for existing files with other format, but they also need an update to their upload key. At the end i decided leave it as it is. After the merge and cname swap is done, I will clean them case by case, and then remove the file_format other.

The insert for txt had the extension txt, changed it to txt.gz because we have existing txt format items with that extension. Again, after merge, we can change file format name to compressed txt.

I added a validator for file_format restriction per file_type. Some of the test were old and not setup to work with file_format item, either fixed them or deleted. Wrote new test for the validator.

The validator addition to files were a bit off, there were no validators for PUT on files, but only for processed files. Changed all validators to only run on File with PATCH PUT POST, and file_processed specific ones will return true on other file types.

Added more file_format inserts to cover all cases on DATA. Also added valid_item_types to all inserts.

PR #872, opened 2018-09-12
Experiment type

sbreiff:


src/encoded/schemas/experiment_type.json
```diff
@@ -0,0 +1,151 @@
...
+            "linkTo": "OntologyTerm"
+        },
+        "experiment_category": {
+            "title": "Experiment Category",
+            "description": "General category of the experiment type",
+            "type": "string",
+            "enum": ["Sequencing", "Microscopy"],
+            "lookup": 100
+        },
+        "assay_classification": {
```
aschroed: For assay_classification and subclassification what about making them suggested_enums?  Makes it significantly easier to add/change the values.

src/encoded/schemas/experiment_type.json
```diff
@@ -0,0 +1,151 @@
...
+            "title": "Other Tags",
+            "description": "Other tags that categorize or describe the experiment type",
+            "type": "array",
+            "items": {
+                "title": "Tag",
+                "description": "A tag for the item.",
+                "type":  "string"
+            }
+        },
+        "raw_files": {
```
aschroed: Maybe raw_file_types would be a better field name?
Again maybe an enum is too restrictive?   Would have to wait through a whole build cycle to add any microscopy raw file types.

src/encoded/schemas/experiment_type.json
```diff
@@ -0,0 +1,151 @@
+{
+    "title": "Experiment Type",
+    "description": "Schema for experiment type",
+    "id": "/profiles/experiment_type.json",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "required": ["title"],
+    "identifyingProperties": ["experiment_name", "uuid", "aliases"],
```
aschroed: If we don't want to support aliases for these then should remove aliases from identifyProperties. 
Alternatively it doesn't hurt to have aliases so could add the mixin.

src/encoded/types/__init__.py
```diff
@@ -96,6 +96,41 @@ class Enzyme(Item):
     embedded_list = Item.embedded_list + ['enzyme_source.title']
 
 
+@collection(
+    name='experiment-types',
+    unique_key='experiment_type:experiment_name',
+    properties={
```
aschroed: I think you might want to add a lookup_key as well for 'title' so that it can be used as the value in spreadsheets for submissions.  Otherwise submitters will need to use the experiment_name form or a uuid (or alias if you decide to add the mixin).

src/encoded/schemas/experiment_type.json
```diff
@@ -0,0 +1,151 @@
...
+            "title": "SOP",
+            "description": "Approved SOP for the experiment type",
+            "type": "string",
+            "linkTo": "Protocol"
+        },
+        "references": {
+            "title": "Reference Publications",
+            "description": "Reference publication(s) for the experiment type",
+            "type": "string"
+        }
```
KorayKirli: We have a depreciated field 'references' that is link to , which we decided to replace with urls. We can do the same here , keeping the title same, but changing the property key (and maybe adding a format = url, ie
```
"url": {
            "title": "Reference Publications/Sources",
            "description": "Reference publication(s) for the experiment type",
            "type": "string",
            "lookup": 1000,
            "format": "uri"
        }```

PR #873, opened 2018-09-13
juicebox_supplementary_files: Adds juicebox on all hic files.

Chadius:
I moved the juicebox link generation code from the ExperimentSetView to the FileEntryBlock. This does mean every FileEntryBlock will use it, but this avoids duplicating code.

- I expanded some of the logic to use more whitespace and more comments, for readability.
- Experiment and ExperimentSet types needed extra fields to track the file type and href link.


src/encoded/types/experiment.py
```diff
@@ -98,7 +98,9 @@ class Experiment(Item):
         "processed_files.file_size",
         "processed_files.upload_key",
         "processed_files.file_format",
-        "processed_files.file_classification"
+        "processed_files.file_classification",
+
+        "other_processed_files.files.href"
```
carlvitzthum: I don't think this embed is necessary? Generally keeping embeds as minimal as possible is good for performance reasons

src/encoded/static/components/browse/components/StackedBlockTable.js
```diff
@@ -853,9 +851,68 @@ export class FileEntryBlock extends React.Component {
...
+        var externalLinkButton;
+        if (fileIsHic) {
+            // Make an external juicebox link.
+            var onClick = function(evt){
+                // If we're on the server side, there is no need to make an external link.
+                if (isServerSide()) return null;
+
+                // Get the href from the page.
+                var pageHref = file.href,
+                    hrefParts = url.parse(pageHref),
```
carlvitzthum: Need to import url package
`import url from 'url';`

src/encoded/types/experiment.py
Chadius: It doesn't look like it from my local testing. I'll remove it and re commit.

PR #874, opened 2018-09-14
Allow additional_info in release updates notes

carlvitzthum:
Added additional_info to data_release_updates secondary items, allowing extra info to be included in the notes section. Added an upgrader, test, and updated inserts and the release-updates page JS

See the yellow release update (Test update 2) for an example. Also, if the additional info is for a secondary item that is the same as the primary item, the additional info is hidden (see green Test update 1)

screen shot 2018-09-14 at 10 59 35 am


src/encoded/static/components/static-pages/ReleaseUpdates.js
```diff
@@ -186,12 +186,15 @@ class SingleUpdate extends React.Component {
...
-        if (secondary_list.length == 1 && set_id === secondary_list[0]['@id']){
+        if (secondary_list.length == 1 && set_id === secondary_list[0]['secondary_id']['@id']){
             return null;
         }else{
-            return secondary_list.map((item) => <div key={item['@id']}><a  href={item['@id']}>{item.display_title}</a></div>);
+            return secondary_list.map((item) => <div key={item['secondary_id']['@id']}>
+                <div>{item.additional_info + ' '}</div>
+                <a href={item['secondary_id']['@id']}>{item.secondary_id.display_title}</a>
+            </div>);
         }
```
alexkb0009: - The `+ ' '` isn't needed unless want non-string values to be converted to strings, e.g. 'True', 'null'.
- Will `item.additional_info` __always__ exist?
  - If not, then could change to `{ item.additional_info ? <div>{ item.additional_info }</div> : null }`
- Minor/Ignorable: Dot notation is preferable in JS to access object properties unless it can't be used for that property due to character (e.g. '@id')
- Should 3 equal sign for vast majority of comparisons (`===` vs `==`) because `==` doesn't compare types.
  - E.g. The following code **will return true** : `1 == true`, `1 == "1"`, etc.
      - Probably ignorable for this case because we are guaranteed to get a number type back from `.length`.

carlvitzthum: Quickly addressing a couple of these points, will tweak code later...
- The `+ ' '` is because I want to have a space character between the additional info (which by convention doesn't include the trailing space) and the `@id` in the href.
- I tried to use dot notation where possible (i.e. no `@id`). However, I went with `item['secondary_id']['@id']` instead of `item.secondary_id['@id']` because it looks weird to me to mix them... I'm happy to switch it if you have a strong preference, though.
- `item.additional_info` _should_ always exist but I think it's best to be as fault tolerant as possible. Good catch.
- Will change to `===`... I definitely remembered that about JS it's just been so long since I've had to write it haha

alexkb0009: For point re: ‘ ‘, the text is set to be rendered within a div element so having a space there would be kinda moot. If want the additional info and link on the same line (with a space) would use a ‘span’ instead.

PR #876, opened 2018-09-18
Tsv download disclaimer

Chadius:
This was more work than I thought it would be, but I'm still learning React.
- I moved the disclaimer HTML into its own class, SelectedFilesDownloadDisclaimer. There's a new button users have to click on to acknowledge they've read the unpublished/unreleased disclaimers. Also disclaimers don't appear if they are unneeded.
- I moved the download button to its own class, SelectedFilesDownloadMetadataButton. I added a close button so you don't have to reach for the x in the corner once you download the files.
- The SelectedFilesDownloadButton class is now the parent container of SelectedFilesDownloadDisclaimer and SelectedFilesDownloadMetadataButton. It determines if the disclaimer is needed in the first place.


carlvitzthum:
The React looks good. My only concern is that the "unpublished data sets" part of the notes is still not emphasized (and appears at the bottom), which may lead to people automatically clicking through the "I've read the notes" button without reading that piece. Thoughts?

carlvitzthum:
Ehh actually ignore that last comment, I didn't realize that the whole "Notes" section is the disclaimer. Maybe rename the notes? I'm cool with it either way

Chadius:
Yeah, I could rename it. Maybe "Before using these files:" or "Attention:" or "For unpublished/unreleased files:"

What are the two hardest things in programming? Refactoring and Naming things, right?

carlvitzthum:
Haha definitely. Regardless, I would let Alex give it a look before you settle on something and merge.

Chadius:
Based on Friday's feedback, we I had to change how the object looks for published files. I added produced_in_pub to the fieldsToRequest and check to see if it exists.

I tested the fixture files to make sure it works in all situations.

PR #878, opened 2018-09-19
Changed size of external sets count size.

Chadius:
browse 4dn data portal - smalltagcrop
The "5 more sets available in External Data" text needs to be larger. I turned the `<small>` HTML element into a `<span>` and used the text size classes found in `encoded/static/components/scss/encoded/_layout.scss` . Here's how it looks with text-large (which is what I coded):
browse 4dn data portal - textlargecrop

I also tried it with text-medium, if large turns out to be too large.
browse 4dn data portal - textmediumcrop


Chadius:
Whoops, I pushed onto the wrong branch. Reverting.

PR #880, opened 2018-09-20
no_results_refactor: No results page uses buttons to direct user.

Chadius:
If the user searches and they don't find any results, they are redirected to this page. It's not immediately obvious they can search for results in External Data.
browse 4dn data portal - original

I decided to change the suggestions into buttons. The text and the button will remind the user they can search for External Data results.
browse 4dn data portal - only external data

If there are no External Data results, we won't show that button and will emphasize the unset filter button.
browse 4dn data portal - no external data


carlvitzthum:
I like the button approach, but IMO the text next to the buttons looks a bit cluttered and doesn't necessarily add much more info. My suggestion would be to make the "Search all items" into a button and fix all three buttons to be the same width, as well as eliminate the text to the right of each button.

Hiding the "Browse external button" when no external sets are available seems like a good idea.

Thoughts @alexkb0009 ?

Chadius:
Let me get rid of the side text and see what it looks like.
I wanted to emphasize searching external data, so I want that button to be the largest.
Do we want to give the advanced search the same emphasis as the other two? If so I can make it a button as well.

carlvitzthum:
Those suggestions are just based off of personal preference; may be worth passing it buy Alex and Burak. You could emphasize the External Data button by making the text on that one the largest and making it green. I just like things matching, so having all three be buttons of the same width is appealing to me

Chadius:
Based on feedback, I removed the advanced search and Unset Filter buttons. If there is External Data, we'll show a button users can click on to load the new page. If they do not generate results, the page will just read "no results found".

PR #881, opened 2018-09-25
React 15 -> 16 Update; NPM dependency cleanup

alexkb0009:
### Before Merging

- ~Wait for HiGlass 1.3.0 release~ -- **DONE**
- ~Update HiGlass version in package.json~ -- **DONE**
- Test `npm run build` time w/ Travis, test UI on local -- **WAITING**

## Changes

The `<head></head>` area of app.js has been slightly refactored, along with corresponding parts in `inline.js` and `react-middleware.js`, in order to address React v16 warnings during `React.hydrate(..)` which pertain to missing/mismatched elements/tags.

Now, instead of the compiled `inline.js` being stuck in as inline JavaScript, `browser.js` is compiled instead straight into `bundle.js`, and the `<script>` tag to request bundle.js is rendered by React directly. Reason is that inline.js, in its position, would execute immediately upon being loaded within the HTML source code and add another `<script>` tag to request `bundle.js` -- which occurs before `React.hydrate()` call and causes 'losing' a tag or two from `<head>` at best.

Some NPM packages which are no longer used have been removed, e.g. https://github.com/facebook/flow. Flow did not provide a high enough utility-to-effort ratio esp. as it is not strict (as TypeScript it) and Visual Studio Code and similar IDEs provide JavaScript type inference out of the box, especially when ESDoc strings are up-to-date.


src/encoded/static/browser.js
```diff
@@ -70,9 +68,9 @@ if (typeof window !== 'undefined' && window.document && !window.TEST_RUNNER) {
         var server_stats = require('querystring').parse(window.stats_cookie);
         var UseApp = connect(store.mapStateToProps)(App);
         var app;
-        
+
         try {
-            app = ReactDOM.render(<Provider store={store}><UseApp /></Provider>, document);
+            app = ReactDOM.hydrate(<Provider store={store}><UseApp /></Provider>, document);
```
alexkb0009: ReactDOM.render was deprecated in React v16 ( https://reactjs.org/docs/react-dom.html#render )

alexkb0009:
Waiting on HiGlass 1.3.0 Update before can merge this one.

alexkb0009:
HiGlass v1.2.5 is out which has non-minified hglib.js. We should be OK to proceed with testing, reviewing, merging.

PR #889, opened 2018-10-15
(Nearing Readiness) Stats Viz pt.2

alexkb0009:
## Introduces tab/chart(s) for Google Analytics and Download Tracking

And minor edits for faceting and searching.

### Backend Search Updates (`search.py`)
- **Searching**
  - Removed code relating to handling of URI params `before` & `after`.
  - Replaced `before` & `after` params with similar code to handle `<field>.to` and `<field>.from` where `<field>` may be any date or numerical field.
   - A _range_ filter is used for any such date or numerical field requested in URI params if it is provided with a `.to` or `.from` prefix, versus a _term_ filter for existing 'term match' facet filters.
   - (**TODO: CHECK**) if improperly-formatted date or number is sent in a .to or .from param, an exception should be thrown somewhere on back-end -- possibly by ElasticSearch itself -- and a 400 error (Bad Request) returned.
   - If multiple 'terms' are selected for a range are sent in URI params, i.e. multiple values for the same `datefield.to` field sent, the back-end will use only one -- preferring the lower value for `.from` and greater value for `.to` in order to expand the single allowed range filter to encompass all/any sent to it.
- **Faceting**
  - Facets with `aggregation_type` == `"range"` or `"date_histogram"` now will be run within "global filters minus self" to get counts for other facet terms if one is selected -- just like existing bucketing (`"aggregation_type" : "terms"`) facets.

### Frontend Search Updates (`FacetList.js`, `experiment-filters.js`)
- Some refactoring which is mostly hidden from sight and not used until we can create a new UI component for 'range facet' . The current design (list of terms) is not friendly or simple to work with in terms of setting range filters.
- Minor refactoring, simplification, & DRYness of click handlers for terms.
- **Searching**
  - Upon selecting a term facet for a range or histogram facet, will generate a `.from` and `.to` filter to add to current URI params.

### Stats Visualization Updates
- Chart for downloads
- Charts for google analytics reports
- UI components to change group by, etc.
- (In Iterative Progress)


src/encoded/static/components/app.js
```diff
@@ -608,25 +608,43 @@ export default class App extends React.Component {
...
-            navigate(href, navOpts, ()=>{
-                var hrefParts = url.parse(href);
-                var hrefHash = hrefParts.hash;
+
+            var navOpts      = {},
+                hrefParts    = url.parse(href),
+                pHrefParts   = url.parse(this.props.href),
+                hrefHash     = hrefParts.hash,
+                targetOffset = target.getAttribute('data-target-offset'),
+                noCache      = target.getAttribute('data-no-cache');
```
alexkb0009: This change allows link elements to have attributes that help inform navigation. E.g. `data-target-offset` would be how far above the target element to scroll to if link href contains a target element ID, e.g. `http://someurl.com/a/page/somewhere#id`. More options can be added as needed, including overrides for navigation options (`navOpts`).

src/encoded/search.py
```diff
@@ -567,94 +554,151 @@ def add_to_sort_dict(requested_sort):
...
+            if query_field not in range_filters:
+                range_filters[query_field] = {}
+                if range_type == 'date':
+                    range_filters[query_field]['format'] = 'yyyy-MM-dd HH:mm'
+
+            if range_direction in ('gt', 'gte', 'lt', 'lte'):
 
-        if not_field:
-            field_filters[query_field]['must_not_terms'].append(term)
+                if len(term) == 10:
```
carlvitzthum: Should this be `if len(term) == 10 and range_type == 'date':`?

PR #898, opened 2018-10-31
HiGlass view: Adding Save, Save As and Share functionality

Chadius:
Adding tests for Save, Save As and Share buttons on HiGlass displays.
- Added Integration tests for Save As, Share and Editing HiGlass displays.
- Added alerts upon completed Save/Save As/Edit actions


src/encoded/static/components/item-pages/HiGlassViewConfigView.js
```diff
@@ -153,32 +160,156 @@ export class HiGlassViewConfigTabView extends React.PureComponent {
...
+                },
+                // We're only updating this object's view conf.
                 JSON.stringify({ 'viewconfig' : currentViewConf })
             );
         });
     }
 
+    /**
+    * Create a new higlass viewconfig for the user, based on the current data.
+    * @returns Nothing
```
alexkb0009: Should be @returns {void}

src/encoded/static/components/item-pages/HiGlassViewConfigView.js
```diff
@@ -119,8 +117,12 @@ export class HiGlassViewConfigTabView extends React.PureComponent {
         return !!(_.findWhere(this.props.context.actions || [], { 'name' : 'edit' }));
     }
 
+    /**
+    * Update the current higlass viewconfig for the user, based on the current data.
+    * Note that this function is throttled in constructor() to prevent someone clicking it like, 100 times within 3 seconds.
+    * @returns Nothing
```
alexkb0009: Should be @returns {void}

src/encoded/static/components/item-pages/HiGlassViewConfigView.js
```diff
@@ -153,32 +160,156 @@ export class HiGlassViewConfigTabView extends React.PureComponent {
...
+                'style' : 'danger'
+            });
+        }
+
+        // Generate a new title and description based on the current display.
+        var userDetails = JWT.getUserDetails();
+        let userFirstName = "unknown";
+        if (userDetails && typeof userDetails.first_name === 'string' && userDetails.first_name.length > 0) userFirstName = userDetails.first_name;
+
+        const viewConfTitle = this.props.context.title + "-" + userFirstName + "'s copy";
```
alexkb0009: This will return` HiGlass Display Name-Bob's copy`. Maybe the `"-"` should instead be `" - "`?

PR #903, opened 2018-11-05
UserContent Base Class; linkTo UserContent/StaticSection/HiglassViewConfig

alexkb0009:


src/encoded/types/user_content.py
```diff
@@ -0,0 +1,161 @@
...
+            else:
+                return 'txt' # Default if no file extension.
+        return None
+
+
+
+
+@collection(
+    name='higlass-view-configs',
+    properties={
```
aschroed: Do you not also need/want a uniquekey here?

src/encoded/types/user_content.py
```diff
@@ -0,0 +1,161 @@
...
+        'title': 'HiGlass Displays',
+        'description': 'Dsiplays and view configurations for HiGlass',
```
aschroed: Typo - Dsiplays -> Displays

src/encoded/types/user_content.py
alexkb0009: It appears to inherit user_content:name from UserContent w/o issue. Could explicitly add `unique_key="user_content:name"` to it though to add clarity. Seem if add `unique_key="static_section:name"` some errors start to be thrown on insert or create-mapping.

alexkb0009:
Closing pull request, instead will put up for review + merge the combined PR of this and the permissions branch.

PR #919, opened 2018-11-20
Add higlass view to config demo

Chadius:
We want to add files to the Higlass Display. This task figures out if the file is 1D or 2D, adds or creates views, and repositions them to share space.

If you go to /higlass-view-configs/ and look at a display, you'll see a new button after Copy URL named Add. Type in the file UUID and click the button to add it to the display.
- Added a React Component to capture the text input and add the file.
- Added addFileToHiglass function to call the underlying Python function and update the Higlass Display.

On the backend, several functions written in Python figure out what the file is, where to place it, and modify the view config to handle it.
- Under the New endpoint, a POST request to /add_files_to_higlass_viewconf/. The JSON payload includes "higlass_viewconfig", The JSON of the viewconfig you want to edit. It also has "files"
, a comma-separated list of file uuids.
- Added unit tests to handle a few scenarios.


Chadius:
Adding more updates for this PR.
- Higlass Viewconfig schema modified so it tracks the genome assembly.
- Updated and improved genome assembly checks so you can't mix mouse and human data.

Chadius:
Added bed file support.
The bed file insert I used did not have Genome Assembly, so I added it directly. I assume any file ready for Higlass will contain a Genome Assembly.

Chadius:
Tests in test_higlass now send a list of files instead of a comma separated string.

PR #938, opened 2018-12-03
higlass-last-second-swaps: Using curated track titles as a filter

Chadius:
Now using the track_title field when presenting files to add to the HiGlass display.
HiGlass also shows the track_title field on each view.

### DIFF
--- a/src/encoded/visualization.py
+++ b/src/encoded/visualization.py
@@ -732,11 +732,8 @@ def add_2d_file_to_higlass_viewconf(views, new_file):
 def get_title(file):
     """ Returns a string containing the title for the view.
     """
-    title = file["display_title"]
-    if "file_classification" in file:
-        title += "({classification})".format(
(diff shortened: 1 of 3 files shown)


src/encoded/visualization.py
```diff
@@ -732,12 +732,11 @@ def add_2d_file_to_higlass_viewconf(views, new_file):
...
     """ Returns a string containing the title for the view.
     """
-    title = file["display_title"]
-    if "file_classification" in file:
-        title += "({classification})".format(
-            classification = file["file_classification"]
-        )
-    return title
+    # Use the track title. As a fallback, use the display title.
+    if "track_title" in file:
```
carlvitzthum: In case `track_title` gets a value of None, this should probably be either:
`if "track_title"` in file and file["track_title"]:`
or equivalently
`if file.get("track_title"):`

PR #955, opened 2018-12-12
Change storage of 'currentAction' from hashbang suffix to URI param

alexkb0009:
Now, instead of navigating to (e.g.) `/search/?type=Experiment#!selection`, we will navigate to `/search/?type=Experiment&currentAction=selection`.

This allows the currentAction to be picked up server-side and be used to inform the server-side render, instead of needing to wait for front-end JS to initialize, components to mount, and URL to be parsed client-side and then viewport re-rendered.

This change is propagated to every single action we have:
- `/<item_@id>/#!edit` -> `/<item_@id>/?currentAction=edit`
- `/<item_@id>/#!create` -> `/<item_@id>/?currentAction=create`
- `/<item_@id>/#!clone` -> `/<item_@id>/?currentAction=clone`
- `.../search/?type=<ItemType>#!add` -> `.../search/?type=<ItemType>&currentAction=add`
- `/#!impersonate-user` -> `/users/<own_uuid>/?currentAction=impersonate-user`


src/encoded/static/components/app.js
```diff
@@ -486,23 +489,25 @@ export default class App extends React.Component {
...
-            var context_actions = [];
-            Array.prototype.push.apply(context_actions, context.actions || []);
+            var context         = this.props.context,
+                currentAction   = this.currentAction(),
+                contextActions = [];
+
+            Array.prototype.push.apply(contextActions, context.actions || []); // Add any from context itself
 
-            if (!name && context.default_page) {
+            // The below is deprecated ? TODO: Double check & delete.
```
Chadius: Are you leaving this if statement behind just in case you missed one behavior?

src/encoded/static/components/facetcharts.js
```diff
@@ -35,6 +35,7 @@ export class FacetCharts extends React.PureComponent {
     static defaultProps = {
         'href' : '/',
         'show' : function(path, search, hash){
+            if (typeof search === 'string' && search.indexOf('currentAction=impersonate-user') > -1) return false;
             if (typeof hash === 'string' && hash.indexOf('!impersonate-user') > -1) return false;
```
Chadius: Both lines say "don't show the FacetCharts if we're trying to impersonate the user." Why do we need both?

src/encoded/static/components/item-pages/WorkflowRunTracingView.js
```diff
@@ -189,14 +175,13 @@ export class FileViewGraphSection extends WorkflowGraphSection {
             'tab'       : <span data-tip={tooltip} className="inline-block"><i className={iconClass} /> Graph</span>,
             'key'       : 'graph-section',
             'disabled'  : !Array.isArray(steps) || steps.length === 0,
-            'isDefault' : isGraphSectionOpen(parentProps.href, hash),
```
Chadius: Can you remove isDefault from the documentation:

`* @returns {{ tab: JSX.Element, key: string, disabled: boolean, isDefault: boolean, content: JSX.Element }} Tab object
`
https://github.com/4dn-dcic/fourfront/pull/955/files#diff-99c88c3cf485556c470772685d051bb3R156

src/encoded/static/components/viz/QuickInfoBar.js
```diff
@@ -33,6 +33,7 @@ export default class QuickInfoBar extends React.PureComponent {
         // If have href, only show for /browse/, /search/, and / & /home
         var urlParts = url.parse(href);
         if (urlParts.hash && urlParts.hash.indexOf('!impersonate-user') > -1) return true;
+        if (urlParts.search && urlParts.search.indexOf('currentAction=impersonate-user') > -1) return true;
```
Chadius: It looks like we're keeping both methods of impersonate-user here - is that "just in case" users can impersonate in 2 different ways?

carlvitzthum: Agreed, seems like `if (urlParts.hash && urlParts.hash.indexOf('!impersonate-user') > -1) return true;` can be removed

src/encoded/static/components/app.js
carlvitzthum: I think the whole `category === 'context'` block can be:
```
if (category === 'context') {
            var context         = this.props.context,
                contextActions = [];

            Array.prototype.push.apply(contextActions, context.actions || []); // Add any from context itself
            return contextActions;
        }
```

src/encoded/static/components/facetcharts.js
carlvitzthum: I would say neither of these lines are necessary, since impersonate user is now done from a /user/ page and not the homepage (facet chart won't be shown there regardless)

src/encoded/static/components/viz/QuickInfoBar.js
carlvitzthum: Same comment as above -- I think both checks for impersonat user can be removed since that is now done from the user page

carlvitzthum:
Another comment: since we now have the query parameter for `currentAction=selection`, could we move the `inSelectionMode             = currentAction === 'selection'` functionality in SearchView.js from render() to the constructor() or something? Seems like that would get rid of the split second where you see the base search view when entering selection mode

src/encoded/static/components/app.js
alexkb0009: No idea what the `default_page` field is or is meant to be so left it in w/ the TODO comment.

src/encoded/static/components/item-pages/WorkflowRunTracingView.js
alexkb0009: The `isDefault` field can still exist on the tab object (it'd make it the default tab) as of now.
Had forgotten to remove this when made adjustment that in e.g. `/experiment-set-replicates/accession/#somestring`, `somestring` was used to refer to a node id of a graph vs now it refers to a tab on an ItemView.

Will double check that leaving `isDefault` in doesn't cause issues with the ability to navigate directly to a tab (`/experiment-set-replicates/accession/#tab_id`) and if does then will remove that feature.

If no issues will update the docstring typedef to be `{{ tab: JSX.Element, key: string, disabled?: boolean, isDefault?: boolean, content: JSX.Element }}` (adding `?`; need to doublecheck esdoc reference) to signify that those are optional.

src/encoded/static/components/app.js
carlvitzthum: @alexkb0009 I don't think we use `default_page` (maybe a holdover from ENCODE?)

Up to you if you want to change; I think removing that code now is a good move for the sake of cleanliness

alexkb0009:
> Another comment: since we now have the query parameter for `currentAction=selection`, could we move the `inSelectionMode = currentAction === 'selection'` functionality in SearchView.js from render() to the constructor() or something? Seems like that would get rid of the split second where you see the base search view when entering selection mode

I would keep it in render method because (devils advocate) `currentAction === 'selection` is a super cheap operation and storing it in a state object might use up more clientside memory/RAM. We definitely could/should not do it in constructor unless we also add:
```
componentWillReceiveProps(nextProps){
    if (nextProps.currentAction === this.props.currentAction){
        this.setState({ inSelectionMode : nextProps.currentAction === 'selection' });
    }
}
```
In order to keep `inSelectionMode` up to date. If we navigate from 1 SearchView to another SearchView (e.g. via user clicking on a facet), the Component instance(s) are never destructed, they just get re-rendered with new prop values so we would need to keep state.currentAction up to date anyway which would negate the benefits of storing/caching `inSelectionMode`.

----

The reason there is a split second where the 'Select' checkboxes aren't visible still is because of the check for `isThereParentWIndow` because that can only be done when mounted client-side. We could remove the check for `isThereParentWindow` but drawback is if someone arrives at selection mode search view _without_ having opened it from a `LinkToSelector` component, the buttons don't do anything. It shouldn't occur so I guess fine to remove probably... though also non-zero possibility.

src/encoded/static/components/item-pages/WorkflowRunTracingView.js
alexkb0009: Yep so according to TabbedView.js, what happens now with isDefault is if is set on any tab object on any Item page, it will be the first tab that will be open when load or navigate to that item page.

If there is a hash as well at the URL, after page load or tabview mounts, it will switch tabs to the tab with id that matches the hash if it exists. The hash id will override the `isDefault` flag, if any hash present in URL (as it should be IMO).

Will update the docstring.

src/encoded/static/components/viz/QuickInfoBar.js
alexkb0009: Yep was just in case of... of I don't know what actually lol.
At the last minute changed URL path of impersonate user form from `/?currentAction=impersonate-user` to `/users/currentuseruuid/?currentAction=impersonate-user` so removed.

PR #959, opened 2019-01-07
Moving column definitions to back-end / schemas

alexkb0009:
# Column Definitions on Front-End

## Previous

Column definitions are defined in schemas (= ideal 'source of truth') and in front-end, in multiple places. This configuration of columns, especially on the front-end, has gotten somewhat messy wherein column definitions are set, overridden, etc. multiple times/places in response to various contexts/rules. Goal was to simplify this.

## Current 
Column definitions for the BrowseView have been moved to schemas and defaults for both BrowseView and SearchView are now set in the `search.py` file instead. The default columns, whose definitions may be overridden in schema, are `display_title`, `@type` (if abstract type), `status`, and `date_created`.

On SearchView and on BrowseView React components, the columns as delivered from back-end (`Object.<Object>` (object containing objects as values)) are extended with settings from the front-end-specific `defaultColumnDefinitionMap` -- a mapping of field names to settings (`Object.<Object>`) in order to become a list of "column definitions" (`{ field: string, title: string, render: function, order: number, minColumnWidth?: number, colTitle?: string, widthMap?: Object.<number> }[]`) and then fed into the common SearchResultTable (& HeadersRow, etc.) component(s). The primary purpose of this mapping is to provide custom rendering of values in the search result table, e.g. proper datetime formatting, creating links to linkedTo Item(s), adding icons, etc.

There remain two relics of the existing 'custom handling of columns on front-end':
- On both SearchView and BrowseView, there is a function which may conditionally clone+extend the `defaultColumnDefinitionMap` into a slightly adjusted mapping before using it to transform/extend the from-backend `columns` into full `columnDefinitions`.
- The `display_title` column definition render method is adjusted for both of the mentioned views conditionally -- if is SearchView and we're in selection mode, then the display_title render method is adjusted to also return a 'select' button. If is BrowseView and selected files props are being fed into it, then 'select files from this experiment set' checkbox gets conditionally rendered in the display_title column.

## Future

From data perspective, we maybe could look at the columns/orderings that are defined and update schemas.

# Full Screen Mode on Search + Browse

## Previous

Clicking the 'expand' button at top right of the BrowseView or SearchView would trigger a function that grabs the DOM Node reference of the <body> element, as well as `div.search-page-container` element, calculate the difference in widths, and then adjust width & margin of `div.search-page-container` to fill up more of the screen. This is different from the newer 'full screen mode' state that can be set/unset via a callback function prop passed down from the `BodyElement` component. Existing setup is not React-ful (exits React context to perform UI adjustment) and is more code/management than necessary.

## Current

We utilize the existing BodyElement component and its full screen state mode. BodyElement sets body className to include `is-full-screen` when full screen state is active and we use it in (S)CSS to adjust SearchView and BrowseView page container(s) appropriately to a larger width. We have somewhat different handling of full screen mode for Item and all other pages (disables vertical scrolling) vs Browse and SearchView, but this is now managed in CSS instead of JS and thus simpler.


Chadius:
Newest changes look good to me. Once you address the other 4 comments I'm ready to approve.

PR #964, opened 2019-01-14
Minor refactoring of Item page tables

alexkb0009:
## Current Changes

- NPM packages updated for React, ReactDOM, ReactTooltip
- More column metadata/definitions moved to back-end from front-end.
- Added more descriptions to properties defined in schemas.
- **Search.py**:
  - Refactoring many search.py functions to accept `request` in places of `types`. (`types = request.registry[TYPES]`).
  - `schema_for_field` now caches/memoizes its return result for duration of a request.
  - `OntologyTerm` and `TrackingItem` are now filtered out of `type=Item` search query. To get them, must request `type=Item&type=TrackingItem` or just `type=TrackingItem`.
  - Descriptions from schema properties are added to column metadata, if not set otherwise, to be used for tooltips on frontend.
    - Removed lookup of descriptions for facets & columns from front-end.
  - The `@type` column is now present if _any_ requested types are abstract (e.g. Item, Experiment, File) instead if _all_ are abstract.
- **Frontend**:
  - _Facets_
    - Move functions related to transforming facets re: front-end state from FacetList to more appropriate places in SearchView and BrowseView to make code more straightforward / simple.
    - Ensure facets are sorted on front-end by `order`.
    - FacetList now takes windowHeight into account to determine how many facets to show as expanded initially; previously was set to show first 12+ terms (belonging to any combination of facets) and collapse remaining ones. Should draw more attention to available facets.
  - Minor code refactoring for readability/performance (doesn't change functionality); removal of (some) old unused functions.
  - Replace instances where width for Item views is calculated from the DOM and replace with `this.getTabWidth()` which now uses `windowWidth` to calculate page container width for performance & logic simplicity.
  - _ItemPageTable[Loader]_
    - Changed this and dependent components to accept list of URLs to load instead of object of incomplete Items in order to cleanup unused/unnecessary code.
    - Sorting of table results should now work correctly - should be in order of URLs passed in, assuming that the URLs match the result `@id`s.
    - Moved ItemPageTable from `../browse/components/` to `../item-pages/components/`.
  - _Workflow Graphs_
    - Pulled out `WorkflowGraphSectionControls` into own file. Refactor minor aspects of it for readability only.
    - Ensure graph is receiving width of tab section to fix glitch where switching tabs to/from/to Graph tab would trigger graph contents scroll to left for small graphs.
  - Static Content Location (`static_content.location`) field no longer supports icon in naming convention. Icons for custom tabs must be defined in dictionary on a front-end, unless value is set to simply `"tab"` wherein the tab will inherit `display_title` and `options.title_icon` of the UserContent for which a tab would be created.

#### ChartDataController
- Works well as-is, updating it might reduce the amount of code, though.
- [Flux](http://fluxxor.com/what-is-flux.html) could likely be used to replace this home-brew code and standardize it, but this would add another code dependency.

#### Updating how the `ref` prop is used in response to future breaking changes in React.
- In React v16.3 the way `refs` work has been updated and now the mechanism of setting `ref="mystring"` and then accessing reference to element via `this.ref.mystring` has been deprecated. [Updated documentation](https://reactjs.org/docs/refs-and-the-dom.html).
- This will be done in a separate branch.

#### Cleanup old/unused CSS
- Separate branch
- We could also attempt to migrate to Sass compiler from Compass (deprecated).

#### Splitting code up to be loaded separately for faster initial page loads.
- See https://reactjs.org/docs/code-splitting.html
- Will be done in separate branch in future, at some point, depending on priority, possibly.
- Requires many changes, including updating webpack as a pre-requisite.
- Why (potentially): https://www.nngroup.com/articles/response-times-3-important-limits/
- Theoretically can get us down to 0.5-1s page load time instead of 2s, but is likely low priority because navigating between pages is faster which is more important.

#### AboveTableControls
- This works but layout & architecture could be improved a bit for maintainability.
- Cleanup is not done here because design has high likelihood of changing, e.g. file type selection controls and mechanism.
- Cleanup should be done after or at the same time as filetype selection filtering mechanism is updated because filetype selection filters, to narrow down search results, would require large breaking changes.


src/encoded/static/components/browse/components/CustomColumnController.js
```diff
@@ -21,33 +23,54 @@ export class CustomColumnController extends React.Component {
...
-    addHiddenColumn(fieldKey){
-        var hiddenCols = this.state.hiddenColumns.slice(0);
-        hiddenCols.push(fieldKey);
-        this.setState({ 'hiddenColumns' : hiddenCols });
+    addHiddenColumn(field){
+        this.setState(function(currState){
+            if (currState.hiddenColumns[field] === true){
+                return null;
+            }
+            var hiddenColumns = _.clone(currState.hiddenColumns);
```
Chadius: Why do you clone objects before editing them? To avoid concurrency/race conditions?

src/encoded/static/components/browse/components/CustomColumnController.js
```diff
@@ -97,50 +98,38 @@ export class CustomColumnSelector extends React.Component {
     }
 
     columnDefinitionsWithHiddenState(){
```
Chadius: Oh, this doesn't filter out visible columns. It adds a "hiddenState" field to each column definition. I thought this function was a filter. Can you add a function comment to clarify that?

src/encoded/static/components/item-pages/FileMicroscopyView.js
```diff
@@ -65,17 +64,12 @@ class FileMicroscopyViewOverview extends React.Component {
     render(){
         var { context, schemas, width, windowWidth } = this.props;
 
-        var setsByKey;
-        var table = null;
+        var setsByKey = expFxn.experimentSetsFromFile(context),
```
Chadius: Do you still need setsByKey? You replaced it with setUrls on line 71. If you don't need it anymore, it looks like you also don't need context.

Chadius: Nevermind, I just saw where it's used.

src/encoded/static/components/browse/components/CustomColumnController.js
alexkb0009: Oops, I merged before seeing comments!
Will do on this branch.

src/encoded/static/components/browse/components/CustomColumnController.js
alexkb0009: In React and functional programming paradigms in general <sup>(1)</sup> it is bad practice (in terms of maintainability / reasoning about the code/system) to mutate the input that is passed in. This is very important with React especially because of the component tree structure -- an input variable may be passed in as a prop to multiple different components and if any one of those components mutates that input variable <sup>(2)</sup> -- without performing an appropriate state update or similar to inform other components of that change -- then will often start to receive unexpected results.

Examples of unexpected results include case where 2+ separate child components might transform the same input variable/prop differently and then pass it down to other components or elements which require different transformed structures -- at least one of those components which requires transformed structure would likely throw an exception. Smaller but harder to pinpoint issues will likely occur such as parent component not updating (because no state update triggered) in response to the data structure being transformed further down the rendering tree by a child component until something else triggers parent component update/rerender which could cause things to unexpectedly change on page (or errors) as parent component attempts to rerender something based off of the transformed-by-child variable (if said variable is a prop of or held in parent component state). And so on.

In our case in particular this is important because the `columns` are originally passed in from `context.columns` of the API endpoint response JSON, before this object of objects (_= object references_ **to same objects in context.columns -- unless they're cloned**) references gets converted into an array of object (references):


This `context` structure is rendered out in the DOM via the `BodyElement` Component:


If the `context` (or any of its properties, via reference) get modified in place, then there is a chance that we will start to get errors in between server-side rendering and client-side rendering as that string will change - perhaps unexpectedly - and differ from the response JSON that was actually output by API endpoint. 


<sup>(1)</sup> React is largely a functional programming paradigm in that components take in input (props) and return output (JSX of HTML to be rendered in DOM).
<sup>(2)</sup> Applicable data structures which can be mutated include: object, array, object/array referenced from object property, object/array as an array element, etc.

Chadius: I didn't think about the situation with 2 child components editing the same object, but that does make sense. Thanks for the explanation.

src/encoded/static/components/browse/components/CustomColumnController.js
Chadius: Thanks!

src/encoded/static/components/item-pages/FileMicroscopyView.js
alexkb0009: I'll maybe try to change func `experimentSetsFromFile` to return a list instead of object keyed by `@id` if is not a big edit.

PR #965, opened 2019-01-16
Viral treatment

sbreiff:
- Added 'Biological' to treatment_type enum
- Added 'biological_agent' field to treatment_agent schema
- Added 'MOI' to concentration_units enum in treatment_agent schema
- Added logic for biological treatment display title
- Added test for biological treatment display title

### DIFF
--- a/src/encoded/types/treatment.py
+++ b/src/encoded/types/treatment.py
@@ -100,7 +100,7 @@ class TreatmentAgent(Treatment):
         if chemical:
             dis_tit = chemical + " treatment" + conditions
         elif biological_agent:
-            dis_tit = biological_agent + " infection" + conditions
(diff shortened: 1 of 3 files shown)


src/encoded/types/treatment.py
```diff
@@ -99,6 +99,8 @@ def display_title(self, request, treatment_type=None, chemical=None,
 
         if chemical:
             dis_tit = chemical + " treatment" + conditions
+        elif biological_agent:
+            dis_tit = biological_agent + " infection" + conditions
```
aschroed: My only concern here is that not all treatment with biological agents may be infections - even though it's a little awkward might be more future proof to just keep it as  Virus treatment?

PR #975, opened 2019-01-24
Updating React code re: v16 changes. Related Improvements.

alexkb0009:
# Summary

These are edits in response to future breaking changes introduced in React 16. These include handling/usage of `refs` as well as the deprecation of the lifecycle methods `componentWillReceiveProps`, `componentWillUpdate`, and `componentWillMount` which are now deemed to be unsafe.

Not all instances of these lifecycle methods have been updated because of the large amount of them in our codebase. We should work to refactor these over time as our work touches upon those components. Many of these changes lead into larger architectural changes and improvements that were or could be done.

A couple of useful docs or posts for going forward with front-end component design:
- [You Probably Don't Need Derived State](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) - June 2018 by [Brian Vaughn](https://github.com/bvaughn)
- [Refs and the DOM](https://reactjs.org/docs/refs-and-the-dom.html)

### NPM Packages added or updated
- **New package** [`memoize-one`](https://www.npmjs.com/package/memoize-one):
  - Is used in blog post referenced above. Added to 4DN to help replace certain instances of `componentWillReceiveProps` which had aimed to reduce JS (re-)computation by caching value(s) which are derived from prop values to Component state.
- **New package** [`react-transition-group`](https://www.npmjs.com/package/react-transition-group):
  - Community-supported project which succeeds the official React [CSS]TransitionGroup component.
  - Provides a simpler mechanism to help manage transitions, especially those of *entering* and *exiting* elements/components within a ListView or similar group in place of attempting to manage own state/refs. See usage in BarPlotChart/ViewContainer.js
- **Updated packages** (in hopes of better React 16/17 compatibility w/ limited success):
  - `nuka-carousel` (4.4.3 -> 4.4.7)
  - `rc-tabs` (7.5.0 -> 9.6.0) 
  - `react-bootstrap` (0.32.1 -> 0.32.4)
  - `react-sticky` (6.0.2 -> 6.0.3)

### Refactoring - React 16/17 Changes
- _(Backend)_ `search.py` no longer adds the `@type` column if in selection mode.
- Instances of string refs -- `<div ref="helloworld"/>` (assignment) & `this.refs.helloworld` (access) -- have been either removed where possible/no-longer-needed or refactored into the current syntax/usage: `this.helloWorldRef = React.createRef()` (constructor), `<div ref={this.helloWorldRef}/>` (assignment), `this.helloWorldRef.current` (access).
  - Most of these edits are straightforward.
- Removing _a few_ cases of `componentWillReceiveProps`
  - Most are still present due to large number of them, as well as expense in refactoring architecturally. We should no longer use `componentWillReceiveProps` for future components and slowly work on migrating current architecture.
  - Where possible, attempted to convert `React.Component` to a `React.PureComponent` to only trigger re-renders upon prop changes. Depending on complexity of component, allows moving `componentWillReceiveProps` logic into the `render` method.
  - In some places, attempted to use new package `memoize-one` in order to cache output of frequently-called _**static**_ functions.
  - In some places, attempted to use `getDerivedStateFromProps` in place of `componentWillReceiveProps`, however this isn't a drop-in replacement and present some challenges:
    - In being a **static** method (whereas `componentWillReceiveProps` is an instance method), the method only accepts `props` parameter in contrast to a `nextProps` which can be compared against `this.props` and thus prevents simple comparison of past vs new props in order to inform updates.
    - Only `componentDidUpdate` receives `pastProps` now to be able to compare against a current `this.props`, and moving logic from componentWillReceiveProps to here would trigger an extra/unnecessary render in almost all cases.
    - There are ways around this -- for example caching a prop into component state to be able to compare/update via `getDerivedStateFromProps` -- but this seems more complex and verbose than it should be.
      - Example - app.js `BodyElement`
      - [Discussion on StackOverflow](https://stackoverflow.com/questions/49723019/compare-with-previous-props-in-getderivedstatefromprops)
    - Alternatively, a snapshot of data can be saved and compared prior to a render (not yet attempted/used).
  - Moved `componentWillReceiveProps` logic from BarPlot Legend component into `componentDidMount` as a re-render is now required/triggered for client-side render anyway in response to receiving a `windowHeight`.

### Refactoring - Performance/UX improvements
- In place of managing own (states of) transitions for BarPlot chart bars, replaced with new NPM package `TransitionGroup` & `CSSTransition` wrapper components. Transitions are now handled in CSS. Results in significant reduction in code and better enter/exit transitions.
- Refactoring `ChartDetailCursor` (and related files) to no longer require/allow a global ~pubsub~ routing mechanism to communicate between `CursorViewBounds` & `ChartDetailCursor` instances. Now, each `CursorViewBounds` renders its own `ChartDetailCursor` instance.
  - React Portal(s) are now used to render the hover element. No effectual change.
  - Memoization and PureComponents have been added in various places to slightly improve (re)rendering performance.
  - There is room to simplify these components still for reusability but very little need/priority at moment.
- For AboveTableControls, removed reliance on `props.windowWidth` and instead added CSS bootstrap classes `hidden-sm` & `hidden-xs` to certain always-rendered elements to hide them at smaller screen sizes.
  - Improves UX on initial page load on smaller browsers. Previously, needed to wait for JS initialization before certain buttons/controls above search result table would resize to an appropriate size.
- For HomePage & other pages which required vertically-centered elements, removed JS "center element vertically" code and replaced with CSS flexbox styling to avoid JS/DOM computation and because HTML5/CSS3 flexboxes are probably mature enough to rely on now. This code is now gone from `utils/layout.js` as well.
- Removed unused FileInput.js (remnant of Encode).
- Memoized certain global static functions which are likely to be called multiple times with the same parameters. Ensured that all parameters passed to memoized functions are often-changed (e.g. not _**same** object reference_ that has potential to mutate).
- For `HiGlassViewConfigView.js`, `hgcApi.exportAsViewConfString()` is now called when `componentWillReceiveProps` detects that URL is changing, which signifies that tab is changing as we store ID of tab as URL hash now. 
  - Likely has room for refactoring/improvement.
  - Is relying still on `componentWillReceiveProps` which we'll need to update out somehow re: React16/17 changes. Perhaps a cached prop/state value (`href`, `isCurrentTab` (=complex) ?) or a snapshot plus `getDerivedStateFromProps` can be used in future.
- Removed some (but not all) code that depended on refs to get element width(s) and explicitly requiring for `width` and `height` to be passed in. This is in response to us now having `windowWidth` and `windowHeight` passed down from `BodyElement` from which we can derive proper-to-use widths from with better performance than accessing DOM elements.
- Removed AJAX call from `LoginMenuItem` component to API `/logout` endpoint as it doesn't do anything of value (we log out by discarding the JSONWebToken).


src/encoded/static/components/app.js
```diff
@@ -1803,26 +1801,6 @@ class BodyElement extends React.PureComponent {
...
 
-    /**
-     * Is executed after ReactTooltip is hidden e.g. via moving cursor away from an element.
-     * Used to unset lingering style.left and style.top property values which may interfere with placement
-     * of the next visible tooltip.
-     *
-     * @private
-     * @returns {void}
-     */
-    onTooltipAfterHide(){
```
alexkb0009: This function/mechanism was removed because newest version of React-Tooltip library has fixed or worked around the issue being addressed here wherein the tooltip would previously become invisible yet linger in its placement/position and prevent subsequent tooltips from intersecting that previous placement/position. Was most noticeable if have a row of closely-spaced buttons and hover over them in order which resulted in inconsistent and unexpected tooltip placement.

src/encoded/static/components/item-pages/components/HiGlass/HiGlassContainer.js
```diff
@@ -917,29 +922,34 @@ export class HiGlassContainer extends React.PureComponent {
      *
      * - this.updateCurrentDomainsInStorage is bound to 'location' change event.
      * - TODO: onDrag/Drop stuff.
+     *
+     * @deprecated
      */
-    bindHiGlassEventHandlers(){
+    bindHiGlassEventHandlers(attemptCount = 0){
```
alexkb0009: This code is only used for FileView, ExpSetView, & ExpView for the _auto-generated_ HiGlassViewConfigs.
It _will_ be removed. Currently this code binds a handler which listens to location changes and then updates location per filetype in browser localStorage to allow user to browse same location of different files.

Further work on this requires UX feedback/planning.

src/encoded/static/components/app.js
```diff
@@ -1471,13 +1476,25 @@ class ContentRenderer extends React.PureComponent {
  */
 class BodyElement extends React.PureComponent {
 
+    static getDerivedStateFromProps(props, state){
+        var stateChange = { 'lastHref' : props.href };
+        if (state.isFullscreen && stateChange.lastHref !== state.lastHref){
```
Chadius: "// Unset full screen if moving away to different pathname."
This comment is still relevant, I would keep it.

Chadius:
Alright, looks good to me besides my one comment about saving a deleted comment.
Normally I don't like huge line changes like this, but that's what happens when you need to refactor, anyway.

src/encoded/static/components/app.js
```diff
@@ -1965,7 +1947,7 @@ class ContentErrorBoundary extends React.Component {
     /**
      * Unsets the error state if we navigate to a different view/href .. which normally should be different ContentView.
      */
-    componentWillReceiveProps(nextProps){
+    componentDidUpdate(nextProps){
```
alexkb0009: Simplest (but not performant) conversion here. Did componentDidUpdate instead of some other update which wouldn't re-render because encountering/having an error ideally would not even happen or be very rare.

PR #958, opened 2018-12-21
Supporting additional Higlass filetypes

Chadius:
primary    | secondary | default location     | track type|accession
----------| -----------|-------------------|-----------|----------
mcool      |           | center               | heatmap|
bigwig     |           | top                  | horizontal-divergent-bar
bedgraph   | bigwig    | top                  | horizontal-divergent-bar| 4DNFIEGXDIZJ
bigbed     |           | top                  | horizontal-vector-heatmap (linear scale, color) | 4DNFIJUMOMQV
bed        | beddb     | top                  | bed-like| 4DNFITFOH1LU
beddb      |           | top, left, searchbar | vertical-gene-annotations, horizontal-gene-annotations, genomePositionSearchBox| 4DNFIH4I1MY4
chromsizes |           | top, left, center    | "horizontal-chromosome-labels", "vertical-chromosome-labels",  "2d-chromosome-grid"| 4DNFIBP173GC

Adding support for multiple filetypes to be added to Higlass displays.
To test adding the chromsize files on your server, revert 61407d2dcaad7a to add a "Add Chromsizes file" button to the higlass config.

I added some additional files to test the new formats and listed them in this table.

PR #978, opened 2019-01-28
higlass-zoom: New 2D views are positioned correctly.

Chadius:
Fixing one of the bugs introduced with https://github.com/4dn-dcic/fourfront/pull/958

When you add another 2D file, the Higlass visualization page will create another view and then keep them in sync, so moving one view moves all of the views. A bug I introduced forgot to set the offset, so the locations did not line up.

Before the bug, adding the same file twice would change their offset.
before

Now they should always be aligned.
after


alexkb0009:
Should the new insert be in `inserts` folder instead?
`master-inserts` inserts get loaded as both test inserts (local) _and_ as production inserts (get added to data.4dnucleome.org).

If don't want em on production and is just for testing, would move to the /inserts/ folder/file.

Chadius:
Good point, let me move these to another folder. We already have this in production, anyway.

Chadius:
Alright, I moved the inserts over.

PR #979, opened 2019-01-28
Update file_processed.json

alexkb0009:


src/encoded/tests/data/inserts/file_processed.json
```diff
@@ -434,7 +434,6 @@
     "genome_assembly": "GRCh38",
     "uuid":"a34d5ea5-eada-4def-a4a7-c227b0d32395",
     "override_experiment_type": "Dataset",
-    "file_type": "File",
```
alexkb0009: Duplicate of `"file_type":"other",` in same insert (?)

PR #984, opened 2019-01-31
Workflow Graphs -- performance Improvements & refactoring

alexkb0009:


src/encoded/static/components/browse/components/table-commons.js
```diff
@@ -39,7 +39,7 @@ export function defaultColumnBlockRenderFxn(result, columnDefinition, props, wid
         }));
     }
 
-    var value = object.getNestedProperty(result, columnDefinition.field);
+    var value = object.getNestedProperty(result, columnDefinition.field, true);
```
alexkb0009: Negligible change here (and below in same file) which makes console output less verbose for when encounter `No value` (or `undefined`) for a search table column.

src/encoded/static/components/item-pages/ExperimentSetView.js
```diff
@@ -116,7 +116,8 @@ export default class ExperimentSetView extends WorkflowRunTracingView {
                 tabs.push(FileViewGraphSection.getTabObject(
                     _.extend({}, this.props, { 'isNodeCurrentContext' : this.isWorkflowNodeCurrentContext }),
                     this.state,
-                    this.handleToggleAllRuns
+                    this.handleToggleAllRuns,
+                    width
```
alexkb0009: Had forgotten to pass this in earlier -- width for provenance graphs.
Allows us to better (re)center workflow graphs in center of screen/viewport when switching between tabs.
Similar change in FileView.js.

src/encoded/static/components/item-pages/WorkflowRunTracingView.js
```diff
@@ -152,12 +152,8 @@ export class FileViewGraphSection extends WorkflowGraphSection {
             'key'       : 'graph-section',
             'disabled'  : !Array.isArray(steps) || steps.length === 0,
             'content'   : (
-                <FileViewGraphSection
-                    {...parentProps}
-                    {..._.pick(parentState, 'steps', 'mounted', 'allRuns')}
-                    key={"graph-for-" + context.uuid}
-                    onToggleAllRuns={onToggleAllRuns}
-                    loading={loadingGraphSteps} />
+                <FileViewGraphSection {...parentProps} {..._.pick(parentState, 'steps', 'mounted', 'allRuns')} width={width}
```
alexkb0009: Width is passed in as a prop now. (Continuation of comment/edit from ExperimentSetView.js.)

src/encoded/static/components/item-pages/WorkflowView.js
```diff
@@ -65,7 +65,7 @@ export function commonGraphPropsFromProps(props){
     };
 
     if (props.isFullscreen) {
-        graphProps.width = props.windowWidth;
+        graphProps.width = props.windowWidth - 40;
```
alexkb0009: There's a 20px padding-left and padding-right on container when in full screen. This accounts for it so we don't have a horizontal scrollbar for the graph container (unless graph width actually exceeds it).

The presence of horizontal scrollbar / graph position on full screen is not 100% consistent at times because the browser seems to return different value for `window.innerWidth` depending on if mouse is attached or not because if mouse is attached, and screen is scrollable, Chrome will add / account-for a vertical scrollbar which is about 15px wide.

src/encoded/static/components/item-pages/components/WorkflowDetailPane/FileDetailBody.js
```diff
@@ -22,108 +23,41 @@ export class FileDetailBody extends React.Component {
...
-        this.maybeLoadFile();
-    }
-
-    componentWillReceiveProps(nextProps){
-        if (nextProps.file !== this.props.file) {
-            this.setState({ file : nextProps.file }, this.maybeLoadFile.bind(this, nextProps.file));
-        }
-    }
-
-    maybeLoadFile(file = this.state.file){
```
alexkb0009: We remove this function as well as all logic relating to it in order to simplify.
We instead add a little bit more info to the File representation in the Workflow/Provenance graph tracing, in place of trying to AJAX file each time.

src/encoded/static/components/item-pages/components/WorkflowNodeElement.js
```diff
@@ -167,16 +167,6 @@ export class WorkflowNodeElement extends React.PureComponent {
         return output; 
     }
 
-    containerStyle(){
```
alexkb0009: This is in CSS now... except for width : columnWidth which isn't needed as width should be inherited from parent element.

src/encoded/static/components/viz/Workflow/Graph.js
```diff
@@ -100,48 +101,12 @@ export default class Graph extends React.Component {
...
-        this.state = {
-            'mounted' : false
-        };
-    }
-
-    componentDidMount(){
-        this.setState({ 'mounted' : true });
-    }
-
-    componentDidUpdate(pastProps){
```
alexkb0009: We removed ability for Graph component to calculate its own width after mounting, in order to simplify.
Width must always be passed in as a prop now.

src/encoded/static/components/viz/Workflow/Graph.js
```diff
@@ -150,126 +115,185 @@ export default class Graph extends React.Component {
...
+
+    /**
+     * Extends each node with X & Y coordinates.
+     *
+     * Converts column placement and position within columns,
+     * along with other chart dimension settings, into X & Y coordinates.
+     *
+     * IMPORTANT:
+     * Returns a new array but _modifies array items in place_.
+     * If need fresh nodes, deep-clone before supplying `props.nodes`.
```
alexkb0009: Will remove line "If need fresh nodes, deep-clone before supplying `props.nodes`."
This is actually wrong --- we cannot deep-clone anything because then object references will be lost.
The input `nodes` and `edges` props must be in same shape / instance as returned from `parseAnalysisSteps`. They **cannot** be deep-cloned in between.

Chadius:
Looks good to me. I tested locally and I didn't find anything odd with the Workflows either.

Chadius:
Hmm, I checked again (on the right branch this time >_>) and noticed some of the edges no longer highlight when you click on them.

For example, the arrow in an md5 file workflow doesn't highlight when you click on the md5.
screen shot 2019-02-01 at 8 38 08 am

screen shot 2019-02-01 at 8 37 57 am

alexkb0009:
Thx @Chadius!
It seems that the report node ought to be disabled in that instance but is not getting correct styling, looking into now.

PR #985, opened 2019-01-31
Quick fix to /release-updates

carlvitzthum:
Update loadUpdates in ReleaseUpdates.js to use field-based search instead of query string

### DIFF
--- a/src/encoded/static/components/static-pages/ReleaseUpdates.js
+++ b/src/encoded/static/components/static-pages/ReleaseUpdates.js
@@ -65,14 +65,17 @@ export default class ReleaseUpdates extends React.Component {
 
 
     loadUpdates(updateTag = null, updateParam = null){
-        var useTag = updateTag || this.state.updateTag || '*';
+        var useTag = updateTag || this.state.updateTag;
         var useParam = updateParam || this.state.updateParam;
-        // enforce date ranges through the query string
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/static-pages/ReleaseUpdates.js
```diff
@@ -65,14 +65,17 @@ export default class ReleaseUpdates extends React.Component {
...
     loadUpdates(updateTag = null, updateParam = null){
-        var useTag = updateTag || this.state.updateTag || '*';
+        var useTag = updateTag || this.state.updateTag;
         var useParam = updateParam || this.state.updateParam;
-        // enforce date ranges through the query string
-        var qString = 'update_tag:' + useTag;
+        // enforce date ranges and required update tag
+        var update_url = '/search/?type=DataReleaseUpdate&limit=all&sort=-date_created';
+        if (useTag){
+            update_url += '&update_tag=' + useTag;
```
alexkb0009: `useTag` and `useParam` could be wrapped in `encodeURIComponent` to be safe. If know that it will never have a character like `#` or `&` it should be fine to leave as-is though.

carlvitzthum: Good call!

PR #986, opened 2019-01-31
Update Node 8->10, Update Gulp 3.9->4

alexkb0009:
### Steps to upgrade (subject to revision)

_All of the following is to be ran from your root fourfront directory._

1. Run `brew update && brew unlink node && brew unlink node@8 && brew install node@10 && brew link node@10 --force`.
   If any errors about NPM files/directories being still being linked or present appear during last command, go ahead and delete `usr/local/bin/npm` and similar files/directories and then repeat `brew link node@10 --force`. (A version of NPM is bundled with Node.)
  `rm /usr/local/bin/npm && rm /usr/local/bin/npx && sudo rm -rf /usr/local/lib/node_modules/npm` should work if needed (thanks @Chadius).

2. At this point you likely need to update your `~/.bash_profile` file to point to node v10.
Homebrew will let you know what to add to it, but alternatively, `export PATH="/usr/local/lib/node/bin:$PATH"` works just as well or better as it'd point to the brew-symlinked version.

3. Restart of terminal you are using to run node/npm may be needed after step 2.

4. Check to make sure the command `node -v` prints `10.15...`.

5. Can then run `npm install -g npm@latest` to update NPM to latest version (6.7.0) if desired.
Then need to perform the following steps on local machine (from github.com/pattern-lab/edition-node-gulp/wiki/Updating-to-Gulp-4 ) -
`npm uninstall gulp -g && npm uninstall gulp && npm install gulpjs/gulp-cli -g && npm install gulp --save-dev`.

6. Finally, reboot the terminal that runs `bin/pserve` (as well as the bin/pserve process, if is running) for changes to take effect for server-side rendering.

7. `npm run dev-quick`, `npm run build`, etc. should now work (again) on local.


Chadius:
```
(venv) TecRef-LT-BE6536:fourfront chad$ brew unlink node
Error: No such keg: /usr/local/Cellar/node
```
I skipped the unlink step, here.

```
sudo rm -rf /usr/local/lib/node_modules/npm
```
I also had to manually delete my npm files, but that may have been an old installation gone wrong.

Does this interfere with fourfront's buildout installer? I just noticed buildout tries to rebuild npm-installer.

alexkb0009:
> Worked for me - thanks! One odd thing is that even though in node_modules the package.json version indicates v6.7 when I run npm -v I still get 6.4.1 even after doing the npm install -g npm. Not sure what is up with that.

Which `node_modules` folder? The one in repo or in like usr/local/lib/node/.. ? Will try to look into.
That command should install a global version of NPM which becomes available in local dir/project w/ terminal restart req'd to take effect.

alexkb0009:
Interesting... is there a `node_modules/npm/package.json` in your fourfront repo that has 6.4.1 by chance? Trying to think about how else a NPM v could override the global 6.7.0 1..

If there is one in local 4dn repo, then can probably just delete it or `npm uninstall npm` (_without_ any `-g` flags) I think..

PR #988, opened 2019-02-04
Additional 3.5s delay before init HiGlass on Static Pages

alexkb0009:
This should be tested on an EC2/AWS environment with real network speeds.

Additionally, contains minor copy-paste-like refactoring to help avoid re-rendering StackedBlockVisual too frequently by no longer passing in new function instance on each render call of VisualBody component.


Chadius:
It looked good on my local machine. You'll want to test it on hotseat or mastertest to get AWS speeds.

PR #990, opened 2019-02-05
Decrease size of Joint Analysis blocks, popover content.

alexkb0009:
Page should now look like the following:
sc2

Should look like the following **(POPOVER ONLY)**:
screenshot

We also have removed some files for fonts/icons which are no longer being used.


src/encoded/static/components/static-pages/placeholders/JointAnalysisMatrix.js
```diff
@@ -300,12 +300,16 @@ class VisualBody extends React.PureComponent {
...
         if (Array.isArray(data)) {
-            return data.length;
+            count = data.length;
         } else if (data) {
-            return 1;
+            count = 1;
         }
-        return 0; // Shouldnt happen?
+        if (count > 100){
+            return <span className="small">{ count }</span>;
```
alexkb0009: Smaller font-size.
We have a few different non-consistent classNames for text sizes that should consider making consistent (maybe from Bootstrap?), including `small` (0.925rem?), `text-small` (0.9rem?), `smaller` & `text-smaller`.

alexkb0009: Replaced with inline styling (0.95rem) which looks a bit better.

src/encoded/static/components/static-pages/components/StackedBlockVisual.js
```diff
@@ -245,30 +246,47 @@ export class StackedBlockVisual extends React.PureComponent {
         return out;
     }
 
-    static aggregateObjectFromList(dataList, keysToShow){
+    static aggregateObjectFromList(dataList, keysToShow, skipParsingKeys=null){
```
alexkb0009: The changes in this function are moot because decided on removing most of the properties from the popover anyway.
What changed is we skip parsing of categorization field (which is used as a title/key) into a JSX/React value, because we need to preserve it as string so it can work as an object key later in process. We also add if condition for handling number values as previously they would go into a list of numbers which then get uniqued into 1+ number that didn't really tell us anything useful (e.g. 'number_of_experiments' showing '2' for a block of 4 expsets that each have 2 number_of_experiments). Numbers now just get summed up from all expsets (if there were any displayed still).

src/encoded/static/components/static-pages/components/StackedBlockVisual.js
```diff
@@ -425,8 +444,8 @@ export class StackedBlockGroupedRow extends React.PureComponent {
 
     static collapsedChildBlocks(data, props){
```
alexkb0009: Nothing has changed in this function except for code formatting.

src/encoded/static/components/static-pages/components/StackedBlockVisual.js
```diff
@@ -643,22 +647,25 @@ export class StackedBlockGroupedRow extends React.PureComponent {
...
-
-        var childBlocks = !isOpen ? this.childBlocksCollapsed() : null;
-
-        var maxBlocksInRow = Math.max.apply(Math.max, _.pluck(_.pluck((childBlocks && childBlocks.props && childBlocks.props.children) || [], 'props'), 'data-block-count'));
+        var childBlocks = !isOpen ? this.childBlocksCollapsed() : null,
+            maxBlocksInRow  = childBlocks && Math.max.apply(Math.max, _.pluck(_.pluck((childBlocks && childBlocks.props && childBlocks.props.children) || [], 'props'), 'data-block-count')),
+            blockHeightFull = blockHeight + blockVerticalSpacing,
+            rowHeight       = blockHeight + (blockVerticalSpacing * 2) + 1;
 
         return (
```
alexkb0009: The output of this render() func is slightly changed in response to styling/CSS edits -- e.g. another element that wraps column label so that label may be centered vertically using inside a flexbox.

src/encoded/static/components/static-pages/placeholders/JointAnalysisMatrix.js
```diff
@@ -136,7 +136,13 @@ export class JointAnalysisMatrix extends React.PureComponent {
 
     static defaultProps = {
         'self_results_url'          : '/browse/?experiments_in_set.biosample.biosource_summary=H1-hESC+%28Tier+1%29&experiments_in_set.biosample.biosource_summary=HFFc6+%28Tier+1%29&experiments_in_set.biosample.biosource_summary=H1-hESC+%28Tier+1%29+differentiated+to+definitive+endoderm&experimentset_type=replicate&type=ExperimentSetReplicate&award.project=4DN&limit=all',
-        'encode_results_url'        : 'https://www.encodeproject.org/search/?type=Experiment&biosample_term_name=H1-hESC&biosample_term_name=HFFc6&status!=archived&status!=revoked&limit=all&field=assay_slims&field=biosample_term_name&field=assay_term_name&field=description&field=lab&field=status',
+        'self_results_url_fields'   : [
```
alexkb0009: These are appended to URL as multiple values for the `field` URI parameter. Back-end then strips all fields from response that aren't present. This is a way to reduce size of AJAXed response without needing to change or re-architect much stuff.

PR #992, opened 2019-02-06
no_more_higlass_io: Removing references to higlass.io

Chadius:
After testing foursight changes, newly created viewconfs were looking for higlass.io to get genome assembly information. They should have been looking at https://higlass.4dnucleome.org/api/v1 .

I've updated all of our inserts, schemas and test data so this stops happening.


Chadius:
Thanks for the catch, Carl. I must have tested using the old inserts (I did not restart my local server after changing the inserts.) I'll do that now.

Chadius:
I have updated the tileset_uids so they refer to similar gene annotations and chromsize files on 4DNucleome, rather than higlass.io (the IDs are different, anyway, as Carl's comment shows.)

I updated some of the initialX/YDomains so the view is centered according to its genome assembly. (For the blank one I just used the assembly with the largest dataspace, GRCh38.)

PR #993, opened 2019-02-08
Higlass remove left side

Chadius:
A few improvements to our HiGlass browser:
1) When composing views from Foursight, if all of the views are 1D then the left tracks are removed. (Pass in "remove_unneeded_tracks = true" when calling add_files_to_higlass_viewconf. Foursight code still needs to be updated first.)

2) If you add a 2D file to a 1D view, all of the reference tracks will be copied from the top to the left side.

3) If you add a 2D file, the view will start centered and the file should take up most of the view. I hardcoded some zoom levels, based on the genome assembly.

empty higlass viewconfig 4dn data portal


src/encoded/visualization.py
```diff
@@ -1144,3 +1198,91 @@ def add_zoom_lock_if_needed(view_config, view, scales_and_center_k):
...
+
+    # Search all views' central contents for any 2D files.
+    for view in new_views:
+        for center_track in view["tracks"]["center"]:
+            if "contents" not in center_track:
+                continue
+
+            # If 2D files are found, we shouldn't remove any tracks.
+            if any([ t for t in center_track["contents"] if t["type"] in ("heatmap", "2d-chromosome-grid")]):
+                return True
```
alexkb0009: Should this be 'False'?

Chadius: Maybe a boolean response is too vague. True meant "no errors found" instead of "I definitely removed the left tracks. I can change it without trouble, though.

Chadius:
Oops, let me get rid of those print statements.

src/encoded/visualization.py
alexkb0009: Ah kk makes sense. Might be cool to have it return 'False' and if any error then let it throw an exception or something in case we wanna know in future if function did/didn't delete a track. Idk, have no real preference.

Chadius:
- The uids for the left and top tracks were the same. HiGlass's renderer would only draw one of them as a result. uid for left track now changes from "top-*" to "left-*" when possible (if not possible it changes to a generated uuid.)
- Left track Orientation is changed from "1d-horizontal" to "1d-vertical".
- If there is no central track (or the track is blank) and you add a 2D file, the camera will be centered to show all of the data.
- viewconfs with only one view do not need a lock.
- When you add a new view, the code added references with duplicate uids. Fixed that.

Chadius:
- If you add a gene annotation file, it is always added in front of the 1D files on the top and left sides.
- Fixed a bug where creating a new view would change the order of the left tracks (the reference files were always the first tracks, even if you added a gene annotation file manually)

PR #1001, opened 2019-02-15
Higlass refactor

Chadius:
I refactored most of the Pyhton endpoint in visualization.py. You may want to read the raw visualizations.py file starting from line 433.
The logic is here:
https://hms-dbmi.atlassian.net/wiki/spaces/FOURDNDCIC/pages/643956741/HiGlass+Workspace

add_single_file_to_higlass_viewconf has major work.
I removed a lot of post processing from the main function to simplify logic.
Now logic is driven by reading the file format and looking at the first view to determine where to add the file and what side effects to do.

- Inspect the left and center tracks of the first view
- Call a function based on the file type
- beddb files are added to the top and left tracks in the first position. Genome position search bar is updated.
- chromsize files are added to the top, left and center tracks in the last position (center tracks are added as a 2d-chromosome-grid type)
- Other 1D functions create a new data track

# Inserts (not for production)
- new beddb file
- new chromsizes file
- fixed coordSystem for GRCh38 file

# Updated unit tests
- chromsize grids are added in the correct order
- The "remove_unneeded_tracks" parameter removes tracks on the left side
- Tracks copied from the top to the left side switch "horizontal" -> "vertical" and "height" -> "width"
- beddb tracks are in the first position in a viewconf's left/top track list, while chromosome tracks are always in the last position.
- Error messages changed slightly


Chadius:
I left something out of the refactor (centering new view configs based on the genome assembly.) I'll readd that now.

Chadius:
Restored the ability to center viewconfigs when creating new configs.
Fixed issues with codacy.

src/encoded/tests/data/inserts/file_reference.json
```diff
@@ -126,24 +126,45 @@
...
+   "uuid": "9866d158-da3c-4d9b-96a9-1d59632eabeb",
+   "description": "Full chromosome sizes file for the human genome reference GRCh38",
    "file_classification": "ancillary file",
-   "higlass_uid": "XGrAqZupQb2h5Pjiarf8PA",
+   "aliases": [
+     "4dn-dcic-lab:GRCh38_chromsizes_full"
+   ],
+   "accession": "4DNFIZJB62D1",
+   "filename": "GRCh38.full.chrom.sizes",
+   "higlass_uid": "NThJBk8mSnqaTzeDaWL-kg",
```
carlvitzthum: "higlass_uid": "JXbq7f-GTeq3FJy_ilIomQ" will use the chromsize file without the unplaced contigs (ugly things at the end of chromsizes grid)

Chadius: Thanks Carl.
You can't see the grid here because HiGlass drew it below the heatmap (the 2d-chromosome-grid is the first track listed in the viewconfig you revealed.) I missed this setup, so I'll fix that now.

HiGlass uid has been updated.

PR #1017, opened 2019-03-06
Update inserts cmd

carlvitzthum:
Command to update any given inserts directory.

For an overview of inserts and the update command, see the [README](https://github.com/4dn-dcic/fourfront/blob/update_inserts_cmd/src/encoded/tests/data/README.md) that is located in the inserts parent directory.

The basic use of the command is to update the current items in any directory of inserts. You can also constrain the command to run on certain item types. Once run, it will pull the `frame=raw` results for existing items AND any associated items from the given FF server using `ff_utils.expand_es_metadata`. Finally, the resulting JSON files are written to the target directory, which will update existing items and add any new ones.

The command is best run BEFORE loading the Fourfront DB locally, since otherwise you would have to run `bin/load-data` to get the updated inserts into the running Fourfront instance.

Example:
```
bin/update-inserts-from-server --env data --dest inserts
```

Another use case of this command is to pull inserts from a live server for testing. When doing this, you should probably use `--dest temp-local-inserts`, which is a gitignored inserts directory meant for local use. It is easiest to do this by leveraging the `--from-search` argument, which performs uses `ff_utils.search_metadata` to find uuids and then also pulls them into the inserts.

For example, to pull a handful of experiment sets from fourfront-hotseat and all associated items into the `temp-local-inserts` directory, you could use:

```
bin/update-inserts-from-server --env fourfront-hotseat --dest temp-local-inserts --from-search search/?experiments_in_set.experiment_type=NAD-seq&type=ExperimentSetReplicate
```

There are a few other things to consider:
- We don't want items in `master-inserts` to be duplicated in `inserts`. So, if you're trying to load an insert that already exists but is different in `master-inserts`, the command will explain why and then fail
- If loading items into `inserts` that are present AND up-to-date in `master-inserts`, the command will automatically remove those items from `inserts` and continue with the loading. The program will explain that in the output
- Though `expand_es_metadata`, `rev_linked` items are only added for file-workflow relationships. See [this code](https://github.com/4dn-dcic/utils/blob/719390b232708938605642d2cabd74da45963ecb/dcicutils/ff_utils.py#L587) for details
- You can use the `--ignore-field` to add field names that will be skipped using `expand_es_metadata`. You can provide as many fields as you want by repeated use of this flag (as with `--item-type`)


carlvitzthum:
Re: @alexkb0009 about the dcicutils issue

Currently master branch of `dcicutils` is git cloned on buildout. It's located in `develop/dcicutils` and you would need to pull from master in that directory. Or, you could run the whole `bin/buildout` which should upgrade to the latest master branch (though that would take a lot longer)

Chadius:
Looks good to me. This will come in handy when we need to locally test.

PR #1021, opened 2019-03-08
Have homepage carousel come from initial request

alexkb0009:
### Changes

- Renamed EncodedRoot to FourfrontRoot for semantics.
- Carousel and announcements now come from the back-end via calculated properties on FourfrontRoot.
- Minor refactoring of BrowseView to be more similar to SearchView (memoized functions instead of (deprecated) `componentWillReceiveProps`)
- Minor refactoring of utils/seo.js (FullSite no longer an instantiable React component)
  - Holding off on embedding/rendering [Carousel structured data](https://developers.google.com/search/docs/guides/mark-up-listings) because would only appear on mobile.
- Moved `<div className="container" id="content">` out of App.js and into individual views to allow more flexibility in laying/styling pages out, e.g. to allow full-page-width carousel using proper markup/CSS.

### Things to Ensure
- `StaticSection.section_type` is **not editable** except by admin.


src/encoded/root.py
```diff
@@ -264,26 +264,39 @@ def jsonld_type(self):
...
         return return_list
 
-#    @calculated_property(schema={
-#        "title": "Announcements",
-#        "type": "array"
-#    })
-#    def announcements(self, request):
-#        '''Returns list of latest announcements'''
-#        try:
-#            ## Doesn't go thru permissions correctly (status:draft visible to non-logged-in-users)
```
alexkb0009: This comment is no longer accurate (?)
Tested and permissions work fine; perhaps is due to `category="page"` ?
Regardless, is working as intended and permissions are correctly observed, even with as_user=True instead of email.

src/encoded/schemas/static_section.json
```diff
@@ -37,6 +37,7 @@
             "title": "Section Type",
             "type": "string",
             "description": "What this section is used for. Defaults to 'Page Section'.",
+            "permission" : "import_items",
```
alexkb0009: Should prevent editing of that field unless admin (have permission to mass-import items).

PR #1024, opened 2019-03-12
Set namekey edit

sbreiff:
- & now changed to n (e.g. cut-n-run)
- other punctuation changed to '-'
- vendor, badge atids should remain the same w/ these changes
- also removed unused badge inserts

### DIFF
--- a/src/encoded/types/base.py
+++ b/src/encoded/types/base.py
@@ -148,11 +148,11 @@ def set_namekey_from_title(properties):
     name = None
     if properties.get('title'):
         exclude = set(string.punctuation.replace('-', ''))
-        name = properties['title']
-        name = re.sub(r"&", ' n ', name)
-        name = ''.join(ch for ch in name if ch not in exclude)
-        name = re.sub(r"\s+", '-', name)
-        name = name.lower()
+        name = properties['title'].replace('&', ' n ')
+        #name = re.sub(r"&", ' n ', name)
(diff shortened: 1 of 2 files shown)


src/encoded/types/base.py
```diff
@@ -147,11 +147,10 @@ def get_item_if_you_can(request, value, itype=None):
...
     if properties.get('title'):
-        exclude = set(string.punctuation)
-        name = properties['title']
-        name = ''.join(ch for ch in name if ch not in exclude)
-        name = re.sub(r"\s+", '-', name)
-        name = name.lower()
+        exclude = set(string.punctuation.replace('-', ''))
+        name = properties['title'].replace('&', ' n ')
+        name = ''.join(ch if ch not in exclude and ch != ' ' else '-' for ch in name)
+        name = re.sub(r"[-]+", '-', name).strip('-').lower()
```
aschroed: I think this is fine. 
Since you need to use the re.sub anyway you could consider doing it all in one regex sub like so.
`exclude = string.punctuation.replace('-', '')
  name = properties['title'].replace('&', ' n ')
  regex =  = "[{}\s]+".format(exclude)
  name = re.sub(regex,  '-', name).lower()`

but I'm not sure that is any clearer and it's essentially the same amount of code so it's up to you.

aschroed: For some reason it didn't line break the code - should be 4 lines.

PR #1029, opened 2019-03-14
Higlass ajax loader

Chadius:
When Billy from Bioinformatics looks at Files with HiGlass IDs, he should be able to explore them and clone his own copies of viewconfigs.

Now that we can generate Higlass Items on foursight, it's time to swap out the rendering methods on the frontend.
- For Files, HiGlassAjaxLoadContainer replaces HiGlassContainer. Ajax container expects an @id to a viewconf and will try to load it. If you pass in the viewconf it will just use that. The viewconf is passed to HiGlassPlainContainer (HiGlassContainer also did this)
- FileVIew doesn't need HiGlassFileTabView anymore for specialized loading, so I deleted the file.


src/encoded/static/components/item-pages/FileView.js
```diff
@@ -30,32 +30,30 @@ export default class FileView extends WorkflowRunTracingView {
...
-            isBWFile    = (fileFormat === 'bw' || fileFormat === 'bg'),
-            isBEDDBFile = (fileFormat === 'beddb' || fileFormat === 'bed'),
-            isBIGBEDFile = fileFormat === 'bigbed';
-        return isMcoolFile || isBWFile || isBEDDBFile || isBIGBEDFile;
+    /**
+    * Looks for static Higlass content in the context.
+    * @param {object} context
+    * @return {boolean} true if one static_content section has HiGlass viewconfs.
+    */
+    static hasHiglassItemInStaticContent(context){
```
carlvitzthum: This function doesn't seem to be used?

src/encoded/static/components/item-pages/components/HiGlass/HiGlassAjaxLoadContainer.js
```diff
@@ -0,0 +1,87 @@
...
+    */
+    getFullHiglassItem(){
+        var { higlassItem } = this.props;
+
+        if (!higlassItem) {
+            return;
+        }
+
+        // If the viewconfig was loaded already, use that
+        if ("viewconfig" in higlassItem) {
```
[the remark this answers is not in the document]
alexkb0009: Bleh sorry I thought these comments had gotten added earlier lol. Current `if` clause should always not be hit so good to merge as-is IMO though w/o `if` clause might be lil cleaner. Doesn't matter much.

src/encoded/static/components/item-pages/FileView.js
Chadius: Yeah, I just noticed that when I started working on the next task. I'll delete it.

PR #1038, opened 2019-03-20
HiGlass Items appear on ExpSet ProcessedFile tabs

Chadius:
Similar to https://github.com/4dn-dcic/fourfront/pull/1029, we're replacing the on-the-fly HiGlass generators with HiGlassItem objects.

- EmbeddedHiglassActions has a new prop, showDescription. If it's set to false the Higlass Item description is not rendered and the "Explore Data" is aligned on the right.
- Replacing HiGlassContainer with HiGlassAjaxLoadContainer and an EmbeddedHiglassActions so the user can click on the button to "Explore Data".
- Adjusted heights so the button and the HiGlass display can fit without bumping into the Files Section

#    Experiment Set Processed Files
We don't show the Higlass Item's title or description on the Processed Files page, as there is not enough room. The "Explore Data" button is aligned right on small and large displays.
ExpSet ProcessedFile Big Screen Shot 2019-04-02 at 1 12 02 PM
ExpSet ProcessedFile Small Screen Shot 2019-04-02 at 1 11 13 PM

- getHiglassItemFromProcessedFiles looks for the static section with "tab:processed-files" and extracts the HiGlass item from it.

To test, you will need to patch a static_content section for one Experiment Set:
```javascript
{
  "static_content":[
    {
        "location": "tab:processed-files",
        "content": "00000000-1111-0000-1111-000000000002",
        "description": "auto_generated_higlass_view_config"
    }
  ]
}
```

# Experiment Set other processed files (aka Supplementary files)
Just like the processed files section, we won't show the title or description for the Higlass item.
ExpSet OPF Small Screen Shot 2019-04-02 at 1 14 43 PM
ExpSet OPF Big Screen Shot 2019-04-02 at 1 13 57 PM

- Instead of searching the static_contents, we search other_processed_files for each group's  higlass_view_config.
- Removing some uneeded arguments from HiGlassAdjustableWidthRow.

To test, you can patch one of the other_processed_files groups. Use a list of accessions for the files field, and add a higlass_view_config field with a uuid.

```javascript
{
	"other_processed_files": [{
		"type": "archived",
		"files": [
			"4DNFIMCOOL01"
		],
		"title": "A super awesome fileset with an MCOOL file",
		"description": "Sed ut perspiciatis unde omnis iste natus error sit voluptatem. Nam libero tempore, cum soluta nobis est eligendi optio cumque nihil impedit quo minus id quod maxime placeat. Temporibus autem quibusdam et aut officiis debitis aut rerum necessitatibus saepe eveniet ut et voluptates repudiandae sint et molestiae non recusandae.",
		"higlass_view_config": "00000000-1111-0000-1111-000000000002"
	}]
}
```

# Files should be unchanged
You can still see the Higlass Item description on File pages.
File Screen Shot 2019-04-02 at 1 16 09 PM

# Other changes
1d tracks work with less vertical space. We assumed a height of 600px, but Processed Files and Supplementary Files only have 300px of height. The workaround is to assume 300px for 1d tracks and let the user manually expand them if desired. We don't have to specify a height for 2d tracks, so they scale to the allocated height.


src/encoded/static/components/item-pages/ExperimentSetView.js
```diff
@@ -442,7 +476,8 @@ export class ProcessedFilesStackedTableSection extends React.PureComponent {
         };
 
         if (currentlyVisualizedFiles && currentlyVisualizedFiles.length > 0){
```
alexkb0009: I think we now no longer need to test for `currentlyVisualizedFiles` (nor `files` further downstream in HiGlassAdustableRow etc.) since only presence of higlassItem(s) should matter now (?)

We could put `const higlassItem = ProcessedFilesStackedTableSection.getHiglassItemFromProcessedFiles(context);` up above in place of `const currentlyVisualizedFiles = ProcessedFilesStackedTableSection.findAllFilesToVisualize(context);` and then change from `if (currentlyVisualizedFiles && currentlyVisualizedFiles.length > 0){` to `if (higlassItem && object.itemUtil.atId(higlassItem)){`.

Then we can stop passing in `files` in `hiGlassProps` object, and change `if (!files || !mounted || !higlassItem) {` to `if (!mounted || !higlassItem || object.itemUtil.atId(higlassItem)) {`. (The extra atId check would be just to doublecheck that User has permissions to view it as well, just in case they don't have view permission for whatever reason).

src/encoded/static/components/static-pages/components/BasicStaticSectionBody.js
```diff
@@ -111,11 +111,12 @@ export class ExpandableStaticHeader extends OverviewHeadingContainer {
...
 
     static defaultProps = {
-        'parentComponentType' : BasicUserContentBody
+        'parentComponentType' : BasicUserContentBody,
+        'descriptionprops' :  { 'className' : 'description' },
     };
 
     render(){
-        var { context, parentComponentType } = this.props,
+        var { context, descriptionprops, parentComponentType } = this.props,
```
alexkb0009: Maybe could make this camelCased (`descriptionProps`)(just for consistency).

src/encoded/static/components/item-pages/ExperimentSetView.js
```diff
@@ -6,10 +6,11 @@ import _ from 'underscore';
 import memoize from 'memoize-one';
 import { Collapse, Button } from 'react-bootstrap';
 import { console, object, isServerSide, expFxn, layout, Schemas, fileUtil, typedefs } from './../util';
-import { ItemHeader, FlexibleDescriptionBox, HiGlassContainer, HiGlassPlainContainer, AdjustableDividerRow, OverviewHeadingContainer } from './components';
+import { ItemHeader, FlexibleDescriptionBox, HiGlassAjaxLoadContainer, HiGlassContainer, HiGlassPlainContainer, AdjustableDividerRow, OverviewHeadingContainer } from './components';
```
carlvitzthum: HiGlassContainer is an unused import

src/encoded/static/components/item-pages/ExperimentSetView.js
Chadius: I don't want to remove files from HiGlassAdjustableWidthRow yet, because the Other Processed Files section still relies on it for the collapsedToolTipContent. But I'll remove it from ProcessedFilesStackedTableSection.

Chadius: ProcessedFilesStackedTable uses currentlyVisualizedFiles to display the file table, so I'll keep that. But HiGlassAdustableRow no longer needs it.

src/encoded/static/components/item-pages/ExperimentSetView.js
```diff
@@ -312,7 +312,7 @@ export class HiGlassAdjustableWidthRow extends React.PureComponent {
             leftPanelDefaultCollapsed, leftPanelCollapseHeight, leftPanelCollapseWidth, higlassItem } = this.props;
 
         // Don't render the HiGlass view if it isn't mounted yet or there is nothing to display.
-        if (!files || !mounted || !higlassItem) {
+        if (!mounted || !(higlassItem || files) ) {
```
alexkb0009: Change to:
`if (!mounted || !higlassItem || !object.itemUtil.atId(higlassItem)) {`

We could probably remove `files` from like `propTypes` as well and any other references/mentions of it in this component

alexkb0009: Then again there's also `collapsedToolTipContent()` that just noticed...

Thinking we could remove the func `collapsedToolTipContent()` entirely and then in render method, change `data-html data-place="right" data-tip={this.collapsedToolTipContent()}>` to be `... data-tip="Open HiGlass Visualization for file(s)">` :-P

PR #1044, opened 2019-03-27
anon_higlass_buttons: Anonymous users can see disabled HiGlass buttons with toolips.

Chadius:
If you visit a HiGlass view config page without logging in, you do not see the Save or Clone buttons. This task reveals the buttons as disabled and then presents a tooltip to log in to enable them.
Before on data
On data

After on local
On my dev machine

- Save and Clone buttons appear.
- Clone button is disabled if there is no user session.
- Disabled buttons will never activate tooltips, so the containing element CollapsibleItemViewButtonToolbar will render tooltips if there is no user session. The screen size changes the implementation in (mobile vs desktop).
- Because the CollapsibleItemViewButtonToolbar takes extra time to load, the tool tips are not built immediately. You can see this in production by navigating to any Higlass Display from the higlass menu (don't just refresh the page.) To fix this issue, I added a debounce call to App that will try to rebuild the tooltips 500ms after the last tooltip build. See app.js for those changes.

I noticed if you change the browser width to add or remove the menu, the tooltips may break. You can do this on data. I don't think users will change their browser widths frequently so I didn't change it.

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -428,8 +428,6 @@ export default class App extends React.Component {
             analytics.registerPageView(this.props.href, this.props.context);
 
             // We need to rebuild tooltips after navigation to a different page.
-            ReactTooltip.rebuild();
-
             // Add a debounce so it runs again after a delay, so other components get a chance to mount.
             App.debouncedOnNavigationTooltipRebuild();
         }
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -90,8 +90,6 @@ class Timeout {
     }
 }
 
-
-
 /**
  * The root and top-most React component for our application.
  * This is wrapped by a Redux store and then rendered by either the server-side
@@ -191,6 +189,8 @@ export default class App extends React.Component {
     static defaultProps = {
         'sessionMayBeSet' : null
     };
+
+    static debouncedOnNavigationTooltipRebuild = _.debounce(ReactTooltip.rebuild, 500);
 
     /**
      * Does some initialization, checks if browser HistoryAPI is supported,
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/app.js
```diff
@@ -430,6 +430,8 @@ export default class App extends React.Component {
             // We need to rebuild tooltips after navigation to a different page.
             ReactTooltip.rebuild();
```
alexkb0009: Remove `ReactTooltip.rebuild();` call since `App.debouncedOnNavigationTooltipRebuild();` below should handle it (no point doing 2x in rapid succession).

src/encoded/static/components/item-pages/components/CollapsibleItemViewButtonToolbar.js
```diff
@@ -56,23 +54,29 @@ export class CollapsibleItemViewButtonToolbar extends React.PureComponent {
...
 
-        var { children, windowWidth, collapseButtonTitle } = this.props,
+        var { children, windowWidth, collapseButtonTitle, session } = this.props,
             gridState       = this.state.mounted && layout.responsiveGridState(windowWidth),
             isMobileSize    = gridState && gridState !== 'lg',
             isOpen          = !isMobileSize || this.state.open;
 
+        // If the user isn't logged in, add a tooltip reminding them to log in.
+        var tooltip = null;
+        if (!session) {
```
alexkb0009: CollapsibleItemViewButtonToolbar.js is meant to be re-usable and not tied to HiGlass specifically so that we could re-use it on other Item views that might need a toolbar in their title area later on, e.g. maybe WorkflowRuns at some point.

Would move this `if (!session){...` logic up into `HiGlassViewConfigTabView` and instead have `CollapsibleItemViewButtonToolbar` accept a more-ambiguous `data-tip` (or `tooltip`, w/e name) prop that then gets applied to the `Collapse`/`ButtonToolbar`.

PR #1045, opened 2019-03-27
Joint Analysis - Allow customization via JSX-in-StaticSections

alexkb0009:
### Updates

- We will now be able to pass props into placeholder JSX, e.g.:
  ```javascript
  <JointAnalysisMatrix
      key="someKey" session={session}
      headerColumnsOrder={['H1-hESC', 'H1-DE', 'HFFc6']}
  />
  ```
  _See test insert for changes_
- Instead of `options.filetype`=`"txt"` + `placeholder: ` prefix, now can just set `options.filetype` to be `"jsx"` without prepending 'placeholder:' in the body field.
- This means that JSX can be used now, so may wrap in a `<div>` or other elements if needed. No JavaScript code itself is supported [(minus a few expressions)](https://www.npmjs.com/package/react-jsx-parser), for security.

### Important Changes/Side-Effects

- `key` prop will now always be required; if it is omitted, a new instance of the component is created every time the page/window resizes or literally anything else changes -- instead of just change in prop value(s).
- Props which are passed down from App (or otherwhere) can be used in placeholder Components, however must define them in the body field JSX. E.g. `session={session}` must be explicitly present otherwise the session prop does not get passed into JointAnalysisMatrix and results are not refreshed upon login/out.


alexkb0009:
Might need to update a NPM test, checking out now..

PR #1052, opened 2019-04-01
Delete fields tweak

carlvitzthum:
Meant to go with [this snovault branch](https://github.com/4dn-dcic/snovault/pull/95), which simplifies delete_fields functionality. This branch mainly contains test changes, but also has a JS tweak to make it so the editing user's lab and award are not added when editing/creating a User object


buildout.cfg
```diff
@@ -33,7 +33,7 @@ auto-checkout = snovault
 		dcicutils
 
 [sources]
-snovault = git https://github.com/4dn-dcic/snovault.git branch=master
+snovault = git https://github.com/4dn-dcic/snovault.git branch=delete_fields_validation
```
alexkb0009: Extra reminder to revert at some pt :-P

PR #1054, opened 2019-04-03
Update linking from TOC entry to markdown header

alexkb0009:
+ Minor refactoring

### DIFF
--- a/deploy/post_deploy_testing/cypress/integration/05_browse_views_files_selection.js
+++ b/deploy/post_deploy_testing/cypress/integration/05_browse_views_files_selection.js
@@ -59,7 +59,7 @@ describe('Browse Views - Files Selection', function () {
                 .end();
         });
 
-        it('Can change to sort by date_sorted -- ascending', function(){
+        it('Can change to sort by date_created -- ascending', function(){
             cy.scrollTo(0, 500)
                 .get('.search-headers-row .search-headers-column-block[data-field="date_created"] .column-sort-icon')
                 .scrollIntoView()
@@ -121,7 +121,7 @@ describe('Browse Views - Files Selection', function () {
                                 cy.wrap($toggleFilesOpenButton).scrollToCenterElement().click({ 'force' : true }).wait(300).end()
                                 .get('.search-results-container .search-result-row.open .result-table-detail-container.open .files-tables-container .stacked-block-table input[type="checkbox"]').each(checkUncheckFileCheckbox.bind(this, origSelectedCount)).end()
                                 .wrap($toggleFilesOpenButton).scrollToCenterElement().click({ 'force' : true }).wait(300).end();
-                            });    
+                            });
                         }).end().wrap($toggleDetailButton).scrollToCenterElement().click({ 'force' : true }).end();
                 });
             });
--- a/src/encoded/static/components/static-pages/components/TableOfContents.js
+++ b/src/encoded/static/components/static-pages/components/TableOfContents.js
@@ -3,20 +3,20 @@
 import React from 'react';
 import * as d3 from 'd3';
 import _ from 'underscore';
+import memoize from 'memoize-one';
 import { Collapse, Button } from 'react-bootstrap';
-import * as globals from './../../globals';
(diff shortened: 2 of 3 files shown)


deploy/post_deploy_testing/cypress/integration/05_browse_views_files_selection.js
```diff
@@ -59,7 +59,7 @@ describe('Browse Views - Files Selection', function () {
                 .end();
         });
 
-        it('Can change to sort by date_sorted -- ascending', function(){
+        it('Can change to sort by date_created -- ascending', function(){
```
alexkb0009: Was a typo from while ago I think

src/encoded/static/components/static-pages/components/TableOfContents.js
```diff
@@ -3,20 +3,20 @@
...
-import * as globals from './../../globals';
 import { getElementTop, animateScrollTo, getScrollingOuterElement, getPageVerticalScrollPosition } from './../../util/layout';
 import { isServerSide, console, navigate, object } from './../../util';
 
 
-class TableEntry extends React.Component {
+class TableEntry extends React.PureComponent {
 
-    static getChildHeaders(content, maxHeaderDepth, currentDepth){
+    static getChildHeaders = memoize(function(content, maxHeaderDepth, currentDepth){
```
alexkb0009: On 2nd thought, this memoization isn't really needed here because static fxns are getting called by multiple different table of content entries so is never really memoized.

Memoization could be made working/effective maybe by hosting an instance of a memoized func in each table of contents entry component instance but is more refactoring that doesn't add much benefit/performance so prly not worth doing.

src/encoded/static/components/static-pages/components/TableOfContents.js
```diff
@@ -183,7 +187,7 @@ class TableEntry extends React.Component {
                 (this.props.className ? ' ' + this.props.className : '') +
                 (depth === 0 ? ' top' : '') +
                 (active ? ' active' : '')
-            } data-depth={depth} data-recursion-depth={recurDepth} ref="listItem">
```
alexkb0009: Was going to refactor these out to be React.createRef() instances but these aren't used/consumed by anything so... removed instead.

src/encoded/static/components/static-pages/components/TableOfContents.js
```diff
@@ -201,7 +205,7 @@ class TableEntry extends React.Component {
 
 class TableEntryChildren extends React.Component {
 
-    static getHeadersFromContent(jsxContent, maxHeaderDepth, currentDepth){
+    static getHeadersFromContent = memoize(function(jsxContent, maxHeaderDepth, currentDepth){
```
alexkb0009: Memoization likely not needed here as well

src/encoded/static/components/static-pages/components/TableOfContents.js
```diff
@@ -267,20 +271,24 @@ class TableEntryChildren extends React.Component {
...
-                    if (hAttributes.id){
-                        link = hAttributes.id;
-                    }
+                    linkTitle = linkTitle.replace(hAttributes.matchedString, '').trim();
                 }
-                if (!link) link = TableOfContents.slugify(hString);
+
+                /** @deprecated */
+                if (!link) link = TableOfContents.slugify(linkTitle); // Fallback -- attempt to not use -- may fail.
+
```
alexkb0009: This block of code above is the primary 'fix'. It uses `h.props.id` instead of `link = TableOfContents.slugify(linkTitle)`.

`hAttributes.id` is an exceptional case where someone could define `{#someID.text-400}` below their Markdown heading and then `someID` gets parsed out as hAttributes.id to override header ID manually in content.

src/encoded/static/components/static-pages/components/TableOfContents.js
```diff
@@ -617,44 +622,41 @@ export class TableOfContents extends React.Component {
         );
 
         var marginTop = 0; // Account for test warning
-        if (windowInnerWidth){
```
alexkb0009: Some refactoring - we get windowWidth and windowHeight passed down from App.js now so no need to access window API here.

PR #1059, opened 2019-04-05
SearchResultTable - use `getDerivedStateFromProps` in place of `key`

alexkb0009:


src/encoded/static/components/browse/components/SearchResultTable.js
```diff
@@ -372,28 +359,15 @@ class LoadMoreAsYouScroll extends React.PureComponent {
                     });
                 }
             } else {
-                this.setState({  'isLoading' : false, }, () => this.props.setResults(this.props.results));
```
alexkb0009: This was removed because setResults just sets props.results.... unsure why had that in there originally..

PR #1060, opened 2019-04-08
Bcc2array

aschroed:


aschroed:
> Tried it on local, WAD
> Would be nice to test the upgrade on hotseat
> 
Yes it would be nice although I'm not sure it's necessary as we've done this sort of thing before and hotseat will be occupied with the experiment type and then biofeature upgrades which are more complex.  If it turns out I need to re-deploy again soon I will merge in this branch.

> A Biosample now can accept multiples of biosource, modification, treatment, and bcc.
> I am not sure how we map all these fields if there are multiples of them.
> Do we make some assumptions like these?'

These are good points.  I don't think we have any multiple Biosources as yet do we?  
If so or if we get them we 
> 
> * if there are treatments and modifications they apply to all biosources

I don't think the above would work at least in terms of modifications but not sure - I guess in theory one could dump a bunch of siRNAs or crispr reagents on mixtures of cells?  

> * if there are multiple bcc, we expect a single biosource

Yes I think we can enforce the above.
> * ...
> 
> fourfront/src/encoded/static/components/testdata/browse/context.js
> doc contains reference to string format bcc. Not sure if used anywhere.

I changed this and pushed.  Let's see if any tests fail.

PR #1061, opened 2019-04-10
Add '.display_title' to objects in search query

carlvitzthum:
We want to automatically add `".display_title"` to any object in a search query without a terminal field specified. For example:
```
/search/?type=File&file_format=fastq  -->  /search/?type=File&file_format.display_title=fastq
/search/?type=Experiment&biosample=<acc> --> /search/?type=Experiment&biosample.display_title=<acc>
```

To do this, I leveraged the `schema_for_field` function to find if a queried field schema is `linkTo`. As an aside, this schema information is cached under `request._field_schema_cache`. After adjusting query params as needed, I then store the results as an attribute on request: `request.normalized_params`. This is a webob.MultiDict, just like the `request.params`, and is used for the rest of search in it's stead. This will allow us to more easily alter query parameters within `search` if we need in the future.

For the facet added by the `display_title` normalized parameter, I use the second to last field in the split parameter, which is the linkTo field that we appended display title to. [See here](https://github.com/4dn-dcic/fourfront/pull/1061/files#diff-cd450a147c2935c7a9fded2c64d2231dR871) If the title of the previously mentioned linkTo field is already being used as a facet, append `" (Title)"` to the facet title. [See here](https://github.com/4dn-dcic/fourfront/pull/1061/files#diff-cd450a147c2935c7a9fded2c64d2231dR895). There is [a test](https://github.com/4dn-dcic/fourfront/pull/1061/files#diff-5d8ddbfe1588dab36a07516a77663cabR403) for this.

I also refactored the search a bit by moving `set_doc_types` to the top of `search`. These doc_types are now used when generating the normalized query, whereas before they were being calculated twice in a similar way.

Added documentation and tests.


src/encoded/search.py
```diff
@@ -280,25 +283,70 @@ def execute_search_for_all_results(search):
...
+        while field_schema and ('items' in field_schema or 'properties' in field_schema):
+            try:
+                field_schema = field_schema['items']
+            except KeyError:
+                pass
+            try:
+                field_schema = field_schema['properties']
+            except KeyError:
+                pass
+        if key == 'type' and val in types:
```
alexkb0009: Maybe this could be at top of function, before `field_schema = ...` is calculated?
Since neither the if check nor its return val require the field_schema to yet be calculated, and `type=` is a URI param which would almost always be present, it might make sense performance-wise.

carlvitzthum: Sure thing. I think the performance increase is negligible but would probably be cleaner

PR #1066, opened 2019-04-18
File StackedBlockTables Refactoring, File Grouping, & Layouting/Styling

alexkb0009:
# Changes

## Refactoring

- Separated concerns/logic of StackedBlockTable and File blocks/columns somewhat so that all extra file column rendering is defined in optional `render` functions defined in columnHeader objects, instead of having this be defined/hard-coded in places like `FileEntryBlock`.

- RawFilesStackedBlockTable (a composed StackedBlockTable) and its derivatives such as RawFilesStackBlockTableExtendedColumns now accept a single `experimentSet` prop instead of various properties from `experimentSet`.

- Removed logic regarding changing background of selected StackedBlocks and StackedBlockLists from gradient to flat color during transition upon clicking 'show x more files' (& similar) due to complexity and anti-patterns. Transition will now occur only if small amount of children to collapse/show (without gradient-to-flat-color change), else no transition occurs.

- Removed older CSS & JS which related to, for the most part:
  - ENCODE styling from _tables.scss.
  - File Pair JS logic (replaced by more ambiguous grouping algo)

- *(not done yet / non-functional)* - FilePairBlock component should be renamed to FileGroupBlock. This change will be bundled into subsequent branch.

## File Grouping Logic

Code which was specific to just pairs (files with paired_end) has been removed and replaced with code that ambiguously groups all files together via their `related_files.file` links. New algorithm can handle both bidirectional and unidirectional links.

Whether a group of files is a "file pair" or not is now determined, in context of RawFilesStackedBlockTable, by checking whether all files in said group have `related_files.relationship_type === "paired with"`.

**Q**: Is this appropriate?

## StackedBlockTable Layouting

- The StackedBlockTables should now allow _any_ number of files in a group -- regardless of `relationship_type` -- and allow customization on a per- FileEntry or FileGroup block as to whether include or exclude the checkbox. Manual Example:

- Building on this, "file pairs" groups will get a single checkbox for all its grouped files, while for other groups/cases there will be ___both___ a group checkbox and an individual file checkbox. *Fake example using pairs:*

- "Buttons" in a File row column have been separated out into a new Component (part of aforementioned refactoring) and in order to take up less width, are now presented as options within a DropDown menu.

# Extraneous

## BrowseView & SearchView Fixes/Improvements

- Clicking on the arrows at the side which horizontally scroll the table now do not create a lag for horizontal position of Detail Panes if any are open. [This no longer occurs](https://gyazo.com/d65081002734f31eb219e67ec4848bd1).
- "Columns" dropdown button changed from having gear icon + text to having an icon only in order to fit on medium screen sizes when have a selected files count in Download Button.


src/encoded/static/components/browse/components/AboveTableControls/SelectedFilesControls.js
```diff
@@ -17,18 +17,6 @@ import { wrapInAboveTablePanel } from './wrapInAboveTablePanel';
 var { Item } = typedefs;
 
 
-export class SelectedFilesOverview extends React.Component {
```
alexkb0009: Is not being used anywhere, likely a left-over artifact from before we had Download Buttons / Metadata TSV implemented.

src/encoded/static/components/browse/components/SearchResultTable.js
```diff
@@ -118,18 +116,16 @@ class ResultDetail extends React.PureComponent{
     render(){
         var { open, rowNumber, result, tableContainerWidth, tableContainerScrollLeft, renderDetailPane, toggleDetailOpen } = this.props;
         return (
-            <div className={"result-table-detail-container" + (open || this.state.closing ? ' open' : ' closed')}>
+            <div className={"result-table-detail-container detail-" + (open || this.state.closing ? 'open' : 'closed')}>
```
alexkb0009: Renaming "open" to "detail-open" className here and in SCSS because `.open button.dropdown-button` is a CSS selector for a Bootstrap style for when dropdown is open which gets erroneously activated otherwise.

src/encoded/static/components/browse/components/SearchResultTable.js
```diff
@@ -514,34 +518,34 @@ class ShadowBorderLayer extends React.Component {
...
             </div>
         );
     }
 
-    performScrollAction(direction = "right", depth = 0){
-        vizUtil.requestAnimationFrame(()=>{
-            var change = (direction === 'right' ? 1 : -1) * this.props.horizontalScrollRateOnEdgeButton;
-            var maxScrollLeft = this.props.fullRowWidth - this.props.tableContainerWidth;
-            var leftOffset = this.props.innerContainerElem.scrollLeft = Math.max(0, Math.min(maxScrollLeft, this.props.innerContainerElem.scrollLeft + change));
-            var detailPanes = DimensioningContainer.findDetailPaneElements();
```
alexkb0009: Not only is manually finding these detail panes from the DOM here not performant (esp. when done as part of a non-throttled event callback), but the `setDetailPanesLeftOffset` function on next line had also used its own `requestAnimationFrame` call which thus was adding the delay.

`setContainerScrollLeft` now updates the `state,tableContainerScrollLeft` which is held higher in state and which also gets passed down to detailPanes' left offsets.

src/encoded/static/components/browse/components/file-tables.js
```diff
@@ -4,11 +4,238 @@ import React from 'react';
...
+        const file = this.props.file;
+        const fileFormat = fileUtil.getFileFormatStr(file);
+        return (file && file.href && (
+            // Needs an href + either it needs a file format of 'hic' OR it has a detailed file type that contains 'hic'
+            (fileFormat && fileFormat === 'hic')
+            || (file.file_type_detailed && file.file_type_detailed.indexOf('(hic)') > -1)
+        ));
+    }
+
+    higlassButton(){
```
alexkb0009: This method is defined but actually called/used at the moment.
Should be refactored from button into a separate indicator as before (?).

alexkb0009:
Made minor update to add in HiGlass btn to Processed Files which have a HiGlass section (latest commit).

E.g.:

PR #1079, opened 2019-04-30
ExperimentSetView - File Selection & Download

alexkb0009:
Should look like the following for Processed Files and Raw Files tabs:

### Primary Changes

- All checkboxes are initially checked by default, so that "Download X Files" is initially not disabled and acts as a "Download All" button unless boxes are unchecked manually.
- Processed File table now has a checkbox next to the title in the "File[ Title]" column if `selectedFiles` are passed in as props. It is styled to be aligned with the checkboxes in the individual file rows' title column.
- Raw Files table now has a checkbox in place of the 'download' icon in its 'file-group' column header.

### Secondary Changes
- Modularization of Download button & modal.
- Minor styling adjustments to Modals in response to HiGlassComponent/hglib.css styles interfering with 4DN styles (previously: [Default](https://gyazo.com/a9d44c243540dd8361209d13d2f0eca2) vs when [HiGlass stylesheet present](https://gyazo.com/1c9a07c8607a6be38253ac12db0c4344)), including vertical centering.


src/encoded/static/components/app.js
```diff
@@ -1742,13 +1742,7 @@ class BodyElement extends React.PureComponent {
...
     getViewportDimensions(){
         if (isServerSide()) return;
-
-        var scrollElem      = (window.document && window.document.scrollingElement) || null,
-            windowWidth     = (scrollElem && (scrollElem.clientWidth || scrollElem.offsetWidth)) || window.innerWidth,
-            windowHeight    = (scrollElem && (scrollElem.clientHeight || scrollElem.offsetHeight)) || window.innerHeight;
-            //documentHeight  = (widthElem && widthElem.scrollHeight); Not relevant re: resizing.
-
-        return { windowWidth, windowHeight };
+        return { 'windowWidth' : window.innerWidth , 'windowHeight' : window.innerHeight };
```
alexkb0009: Changed from document.scrollingElement (usually alias of document.body or document html element) to window inner width to align with CSS window-size-based breakpoints.

Downside is some elements/things might appear a little bit wider-than-available-space now on _small_ screens or when full screen because the scrollbar width is included in `window.innerWidth`.

However, (a) small screen sizes occur mostly on mobile where the scrollbar is not rendered and (b) when full-screen, scrollbar is forcibly hidden and window is not usually scrollable (exception: BrowseView, where doesn't matter) so this seemed acceptable.

src/encoded/static/components/browse/components/AboveTableControls/SelectedFilesDownloadButton.js
```diff
@@ -0,0 +1,248 @@
...
+                || (file.from_experiment_set && file.from_experiment_set.produced_in_pub)
+            );
+        });
+    });
+
+    constructor(props){
+        super(props);
+        this.handleAcceptDisclaimer = this.handleAcceptDisclaimer.bind(this);
+        this.state = {
+            'disclaimerAccepted' : false
```
alexkb0009: Cleaned up state. `disclaimerAccepted` is the only state variable that should be needed -- other things/values can be derived directly from it.

Adding more state variables than needed is discouraged as state management is more complex than functional patterns/code.

src/encoded/static/components/browse/components/AboveTableControls/SelectedFilesDownloadButton.js
```diff
@@ -0,0 +1,248 @@
...
+                    </Modal.Title>
+                </Modal.Header>
+
+                <Modal.Body>
+
+                    <p>Please press the &quot;Download&quot; button below to save the metadata TSV file which contains download URLs and other information for the selected files.</p>
+                    <p>Once you have saved the metadata TSV, you may download the files on any machine or server with the following cURL command:</p>
+                    <ModalCodeSnippet filename={suggestedFilename} isSignedIn={isSignedIn} />
+
+                    <div className="extra-notes-section">
```
alexkb0009: Previously this entire section was getting excluded if `!foundUnpublishedFiles` so we moved most things back into a single component for speed/simplicity.

src/encoded/static/components/browse/components/AboveTableControls/SelectedFilesDownloadButton.js
```diff
@@ -0,0 +1,248 @@
...
+    * This function renders out a literal form which may be submitted, with 'accession triples'
+    * ([ExpSetAccession, ExpAccession, FileAccession]) included in the POSTed form fields which
+    * identify the individual files to download.
+    *
+    * @returns {JSX.Element} A modal instance.
+    */
+    render(){
+        const { onClose, suggestedFilename } = this.props;
+        return (
+            <form method="POST" action="/metadata/?type=ExperimentSet&sort=accession" className="inline-block">
```
alexkb0009: Added `className="inline-block"` to form element to allow contents to be rendered in-line, e.g. side-by-side with other externally-placed buttons, removing need for Cancel button to be placed inside the form as well.

src/encoded/static/scss/encoded/_layout.scss
```diff
@@ -583,15 +583,41 @@ $margin-util-sizes-mini: 1, 2, 3, 4, 5, 6, 7, 8; 		// * 1px
 
 .modal-dialog {
```
alexkb0009: Performed minor CSS edits and improved 'targeting' in CSS selector strings so that these styles (including that in `../bootstrap/_modals.scss` above) are prioritized over the overrides [defined in HiGlassComponent/hglib.css](https://github.com/higlass/higlass/blob/a44dcf9226f0e91275ad1ce01f8accf1340a448f/app/styles/TrackOptions.css#L30-L37).

Added [related issue](https://github.com/higlass/higlass/issues/638) to higlass repo.

PR #1085, opened 2019-05-06
Targ to bf update

aschroed:
Still doing some final checking and waiting on a CNAME swap but code should be good to review.


src/encoded/types/modification.py
```diff
@@ -26,7 +26,7 @@ class Modification(Item):
         'constructs.tags',
         'constructs.designed_to_target',
         'modified_regions.aliases',
-        'target_of_mod.target_summary'
+        'target_of_mod.display_title'
```
sbreiff: Is this necessary? My impression is that display_title of a linkTo gets embedded by default, along with @id and @type etc.

src/encoded/types/treatment.py
```diff
@@ -122,7 +122,7 @@ class TreatmentRnai(Treatment):
     embedded_list = Treatment.embedded_list + [
         'rnai_vendor.name',
         'rnai_constructs.designed_to_target',
-        'target.target_summary'
+        'target.display_title'
```
sbreiff: Same question as prev comment

src/encoded/types/experiment.py
```diff
@@ -628,9 +635,11 @@ def experiment_summary(self, request, experiment_type, biosample, targeted_facto
...
 
         if targeted_factor:
-            target_props = request.embed(targeted_factor, '@@object')
-            target_summary = target_props['display_title']
-            sum_str += ('against ' + target_summary)
+            tstring = ''
+            for tf in targeted_factor:
+                target_props = request.embed(tf, '@@object')
+                tstring += ', {}'.format(target_props['display_title'])
+            sum_str += ('against ' + tstring[2:])
```
sbreiff: might need to add a space (e.g. `' against '`) to avoid display titles like `CUT&RUNagainst YFG protein on GM12878`

src/encoded/types/experiment.py
```diff
@@ -670,9 +679,11 @@ def experiment_summary(self, request, experiment_type, biosample, targeted_facto
...
```
sbreiff: `' against '` (see prev comment)

PR #1086, opened 2019-05-07
Stats Page - Hide Range Queries by Default

alexkb0009:
### See [video for demo](https://gyazo.com/3e4b6e06da42545ded7399e268529ddd).

- Slightly [older video also](https://gyazo.com/39753cb85caafe8ba0c10c4feca544af).
- Adds checkbox to toggle between showing range query downloads and not (default off).
- Upon change, chart countby should change appropriately to show range query DLs vs not if not present.

### Extra

- The _**vast**_ majority of this PR is refactoring the `StatisticPageView` and `AreaChart` files to be compatible with React 16.3+ -- which had been put off earlier (and somewhat forgotten) as well as being more Reactful (e.g. using Component composition instead of object/component inheritance), wherein previously the original goal had been to get many charts on a page as quickly as possible.
- This brings some performance degradation in large part due to the deprecation of `componentWillReceiveProps`. Charts are now redrawn an extra time or two now upon, for example, change or reset of color scales (whereas originally it could be batched into a single redraw/update -- along with the data that had changed). This isn't a 'primary page' however so seems acceptable for time being. 
- There are likely ways to refactor this further in order to be performant (& remain simpler/Reactful) but is not yet a priority. Some ideas might be use React to draw out the charts instead of using D3 functions and leverage its virtual DOM diffing. Perhaps to figure out some 'key' value to use on `GroupOfCharts` instead of current 2-step reseting of data and then colorScale.


src/encoded/static/components/static-pages/StatisticsPageView.js
```diff
@@ -708,34 +448,274 @@ export const aggregationsToChartData = {
...
+            untilDate.startOf('month').subtract(1, 'minute'); // Last minute of previous month
+            fromDate = untilDate.clone();
+            fromDate.subtract(12, 'month'); // Go back 12 months
+        } else if (currentGroupBy === 'daily'){ // 30 days
+            untilDate.subtract(1, 'day');
+            fromDate = untilDate.clone();
+            fromDate.subtract(30, 'day'); // Go back 30 days
+        }
+        return { fromDate, untilDate };
+    }
```
alexkb0009: Have tested this func/edit w/ multiple time zones on local.
It might be worth doing untilDate.startOf('day') or something to round better but doesn't quite seem necessary...

PR #1087, opened 2019-05-07
resize-higlass: Build displays with variable heights.

Chadius:
Higlass visualizations have a fixed height of around 300 pixels. This task lets you pass in a variable to control the height.
- 300 pixels is a good height for the Higlass Items on the ExpSet pages. When you look at the Processed Files or Supplementary Files tab all of the 1 dimensional tracks fit.
- Higlass Item page is designed for 600 pixel high displays. We should take advantage of that space.

To prevent 1d tracks from getting too tall, they are capped to 125 pixels in height.

The default height is set to 300 pixels so there should be no change in behavior.

This task will work with future foursight changes to change the heights of some of the Higlass Items.

### DIFF
--- a/src/encoded/tests/test_higlass.py
+++ b/src/encoded/tests/test_higlass.py
@@ -1234,7 +1234,7 @@ def test_add_bigwig_higlass(testapp, higlass_mcool_viewconf, bigwig_file_json):
         AssertionError if the test fails.
     """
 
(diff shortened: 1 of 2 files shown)


Chadius:
`maximum_height` only affects 1d tracks. The default heights are too short.
2d tracks will always expand to fit the given area, so they are unaffected.

PR #1088, opened 2019-05-07
Handle "NOT" in /search/ for linkTos

alexkb0009:
Replicates https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/search.py#L714


alexkb0009:
Cancelled 1 of the Travis builds, hopefully will merge shortly.

PR #1094, opened 2019-05-13
Submissions View Updates

alexkb0009:
- Submitter is loaded
- Submitter can set alias (using own email)
- URL change for check_only=True
- (Some/Unfinished) Refactoring re: linting

For ppl w/ multiple submits_for labs:

For ppl w/ 1 submits_for lab:

People w/ no submits_for labs get the same as w/ 1 but have their email (with `@` replaced with `_at_`)


src/encoded/static/components/forms/SubmissionView.js
```diff
@@ -1072,39 +1081,41 @@ export default class SubmissionView extends React.PureComponent{
...
+                // change actionMethod and destination based on edit/round two
+                if (this.state.roundTwo){
+                    destination = this.state.keyComplete[inKey];
+                    actionMethod = 'PATCH';
+                    const alreadySubmittedContext = this.state.keyContext[destination];
+                    // roundTwo flag set to true for second round
+                    deleteFields = this.buildDeleteFields(finalizedContext, alreadySubmittedContext, currSchema);
+                } else if (this.state.edit && inKey === 0){
+                    destination = object.itemUtil.atId(propContext);
+                    actionMethod = 'PATCH';
```
carlvitzthum: I think we need `deleteFields = this.buildDeleteFields(finalizedContext, propContext, currSchema);` here

carlvitzthum:
Also, another issue I found (haven't looked into where in the code it originates from):

If you have an existing item without an alias (i.e. something from `/search/?type=Item&aliases=No+value`), typing anything in the `aliases` form input causes a bunch of new empty entries in the list to be created. Works fine when prompted for initial alias creation on a new item or editing an existing item with an alias.

carlvitzthum:
Following up on the last comment, seems to be because `aliases` is getting a string value instead of a list in this case. I think this is a pre-existing issue in `master` branch, as I can replicate on mastertest

PR #1097, opened 2019-05-17
Inc type test cov

aschroed:


src/encoded/types/experiment_type.py
```diff
@@ -65,7 +65,7 @@ def assay_subclass_short(self, request, assay_classification=None, assay_subclas
         elif assay_subclassification in subclass_dict:
             return subclass_dict[assay_subclassification]
         else:
-            return assay_subclassification
+            return 'Unclassified'
```
sbreiff: Now that I think about it more, what about using the Category for this? Maybe something like:
```
elif experiment_category:
    return experiment_category + ' (unclassified)'
else:
    return 'Unclassified'
```

PR #1130, opened 2019-07-18
Added prodno to required antibody fields

aschroed:


aschroed:
All current antibody items have been updated on data to have a value in 'antibody_prod_no' field.

PR #1136, opened 2019-08-13
Go ow fix

sbreiff:
Changes to generate_ontology: round 1


src/encoded/commands/generate_ontology.py
```diff
@@ -746,38 +757,41 @@ def main():
...
         terms = add_slim_terms(terms, slim_terms)
         terms = remove_obsoletes_and_unnamed(terms)
-        terms = verify_and_update_ontology(terms, ontologies)
+        terms = verify_and_update_ontology(terms, ontologies, db_terms)
         filter_unchanged = True
         if args.full:
             filter_unchanged = False
-        partitioned_terms = id_post_and_patch(terms, db_terms, ontologies, filter_unchanged)
-        terms2write = add_uuids_and_combine(partitioned_terms)
+        updates, idmap = id_post_and_patch(terms, db_terms, ontologies, filter_unchanged)
```
sbreiff: Do we still want to return the second part of this? It looks like `idmap` is no longer being used outside the function that generates it.

aschroed: You're right this is no longer needed.  Do you want to fix or shall I?

PR #1138, opened 2019-08-20
Fix for non-connecting WFRs/Files

alexkb0009:


src/encoded/types/workflow.py
```diff
@@ -808,19 +802,28 @@ def map_run_data_to_io_arg(step_io_arg, wfr_runtime_inputs, io_type):
...
+                    value_field_name : value_list,
                     "type" : io_type,
-                    "meta" : [ # Aligned-to-file-list list of file metadata
-                        {   # All remaining properties from dict in (e.g.) 'input_files','output_files',etc. list.
-                            k:v for (k,v) in p.items()
-                            if k not in [ 'value', 'value_qc', 'type', 'workflow_argument_name' ]
-                        } for p in matched_runtime_io_data
-                    ]
+                    "meta" : meta_list
                 }
```
alexkb0009: No logic has changed in the above lines, just refactored to make more clear that indices of run_data.meta and run_data.file (or run_data.value, if is parameter and not linkto) are aligned.

PR #1139, opened 2019-08-21
Confirm dialog for HiGlassViewConfigView save button

utku-ozturk:
**Problem:** In HiGlass display item pages, we need to show a warning when user saves on top of an already public higlass-view-conf. this should only show up if the item has status=released or status=released_to_project. (@burakalver)

**Solution:** 
- First of all, a new general-purpose bootstrap modal-popup,`ConfirmModal` component added to ask user's consent for various user interactions e.g. save, delete... (component's title, content, confirm/cancel buttons' text are customizable)
- Then the new `ConfirmModal`component is used in HiGlassViewConfigView in save button's click if the conditions, stated in problem definition above, match.

modal_confirm_dialog


utku-ozturk:
1. Cancel button's css class `btn-default` changed as `btn-outline-warning`. (@alexkb0009 )
2. `mr-08` css rule added to button's to increase gap between icon and text. (@alexkb0009 )
3. Redundant "{}" removed for string-only JSX props. (@alexkb0009)
4. Multiple font styling (italic, bold, ..etc) removed. (@burakalver)
5. "OK" text changed as "Save" 
6. Title and buttons' icon-text alignment untouched since a bootstrap css upgrade failure or higlass css override may cause the problem as @alexkb0009 pointed out. (see the attached image)

** Recent View **
Screen Shot 2019-08-23 at 14 00 25
 
** HiGlass Css Rule Override **
HiGlass Css Rule Overrides Modal Title's Css

PR #1145, opened 2019-08-30
Workflow Edge Routing

alexkb0009:
Now edges should go around nodes, staying closer to Y coordinate of starting node as are closer to it, and closer to Y coordinate of ending node as get closer to it.

A rudimentary/heuristic 'visibility graph' is calculated out of intersection of subdivisions of 'rowSpacing' intervals (spacing between Y coordinates of nodes) and column areas with no nodes occupying it. Subdivisions of rowSpacing are increased and tracing is re-tried (up to 5 times) if can't find/trace a path. Edges then pluck out an available horizontal segment for each column traversed that has Y coord closest to median of previous segment's ending Y coord (or start Y coord, in initial 'pluck') and target Y coord.

This could potentially run into sub-optimal representations that might be correctable by adjusting order of edges traced. Need more test cases.

### DIFF
--- a/src/encoded/static/components/viz/Workflow/EdgesLayer.js
+++ b/src/encoded/static/components/viz/Workflow/EdgesLayer.js
@@ -13,10 +13,14 @@ import Edge from './Edge';
 
 
 
-export function traceEdges(originalEdges, nodes, columnWidth, columnSpacing, rowSpacing, contentWidth, contentHeight, innerMargin){
+export function traceEdges(
+    originalEdges,
+    nodes,
+    columnWidth, columnSpacing, rowSpacing, contentWidth, contentHeight, innerMargin, nodeEdgeLedgeWidths = [10, 10]
+){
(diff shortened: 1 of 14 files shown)


src/encoded/static/components/viz/Workflow/EdgesLayer.js
```diff
@@ -5,11 +5,242 @@ import PropTypes from 'prop-types';
...
+    let horizontalSegments;
+    let tracedEdges = null;
+    let attempts = 0;
+    while (!tracedEdges && attempts < 5){
+        horizontalSegments = buildVisibilityGraph(4 + attempts);
+        try {
+            tracedEdges = assembleSegments(horizontalSegments, 4 + attempts);
+        } catch (e){
+            tracedEdges = null;
+            if (e.message === "Could not find viable path for edge"){
```
alexkb0009: Will include `|| attempts === 4` in if condition & remove from while loop condition

PR #1147, opened 2019-08-30
Fix HiGlass zoom problem after a tab change

utku-ozturk:
**Problem**: HiGlass component loses its state and zoom if page's active tab is changed. To reproduce the bug: adjust the HiGlass zoom then back and forth between tabs.

**Background**: HiGlass component uses a config file, a.k.a `viewConfig`, to initialize its state. The `viewConfig` is implicitly updated while any zoom/pan or misc. user interactions. Component provides an API to import and export its `viewConfig`. Beside this built-in config management, there exists `HiGlassViewConfigView.js/UNSAFE_componentWillReceiveProps` function. It is to save the viewConfig to React state, in order to prevent any lose while user changes the tabs in the page. Theoretically, it is a helpful feature from React perspective. On the other hand, HiGlass components implicit data loading and viewConfig updating is not so transparent, so container React component can hardly determine to save viewConfig. Unfortunately, `HiGlassViewConfigView.js/UNSAFE_componentWillReceiveProps` saves the state even if viewConfig is not stable. (We monitored by logging viewConfig's `InitialXDomain` and `InitialYDomain`, while switch from HiGlass tab, or between any tabs other than the HiGlass'.)

**Solution**: First of all, we limited saving to state: Instead of every tab change, saving runs only while leaving HiGlass container tab. It is trivial to determine HiGlass tab if it is not the first tab: `this.props.href `includes # higlass. On the other hand, if HiGlass tab is the first one, `this.props.href` lacks any # higlass. `rc-tabs` has an `activeKey` attribute, but we couldn't utilize it since cannot propagate of `activeKey` from `TabbedView` to `HiGlassViewConfigTabView` via `rc-tabs`. So we tweaked  `TabbedView` by adding a `hiGlassTabIndex` attribute. This solution worked well, since we can determine whether the recent tab is HiGlass container tab HiGlassViewConfigView.js/UNSAFE_componentWillReceiveProps`. If so save `viewConfig` to state.

While everything seems all set, we noticed that this solution fails for some conditions, e.g. the HiGlass server updated during our tests. After the update, component couldn't retrieve `chrom-sizes.js` file and gets `404` response from server. (see attached image) This condition triggers the component continuously update its `viewConfig`. So we can hardly save a stable version of `viewConfig`. We can hardly trust what we save, so why we need it? Finally, decided to cancel saving to state as once @alexkb0009 stated before. Commented out:

```
//currentViewConf && _.extend(nextState, {
//       'viewConfig' : currentViewConf
//});
```

**PS:** hiGlassTabIndex not used recently, but not removed.

chrom-sizes.js

chrom-sizes-404-error

### DIFF
--- a/src/encoded/static/components/item-pages/HiGlassViewConfigView.js
+++ b/src/encoded/static/components/item-pages/HiGlassViewConfigView.js
@@ -103,39 +103,39 @@ export class HiGlassViewConfigTabView extends React.PureComponent {
      * componentDidUpdate (?) now that we don't swap out state.viewConfig with nextProps.viewConfig.
      * -- After cleanup.
      */
-    UNSAFE_componentWillReceiveProps(nextProps){
-        var nextState = {};
+    // UNSAFE_componentWillReceiveProps(nextProps){
+    //     const nextState = {};
 
-        /*  Below code:
-            We will likely adjust/remove to no longer change viewConfig if receive new one from back-end because
-            backend will always deliver new object reference. Even if same context['@id'] and context.date_modified.
-        */
+    //     /*  Below code:
+    //         We will likely adjust/remove to no longer change viewConfig if receive new one from back-end because
+    //         backend will always deliver new object reference. Even if same context['@id'] and context.date_modified.
+    //     */
 
-        if (nextProps.viewConfig !== this.props.viewConfig){
-            _.extend(nextState, {
-                'originalViewConfig' : null, //object.deepClone(nextProps.viewConfig) // Not currently used.
-                'viewConfig'         : nextProps.viewConfig,
-                'genome_assembly'    : (nextProps.context && nextProps.context.genome_assembly) || this.state.genome_assembly || null
-            });
-        }
-        const hiGlassTabIndex = (this.props.hiGlassTabIndex !== 'undefined') ? this.props.hiGlassTabIndex : -1;        
-        const recentTabIsHiglass = (hiGlassTabIndex === 0 ? this.props.href.indexOf('#') < 0 : false) || (this.props.href.indexOf('#higlass') >= 0);
-        if (recentTabIsHiglass && (nextProps.href !== this.props.href) && (object.itemUtil.atId(nextProps.context) === object.itemUtil.atId(this.props.context))) {
-            // If component is still same instance, then is likely that we're changing
-            // the URI hash as a consequence of changing tabs --or-- reloading current context due to change in session, etc.
-            // Export & save viewConfig from HiGlassComponent internal state to our own to preserve contents.
-            var hgc = this.getHiGlassComponent(),
-                currentViewConfStr = hgc && hgc.api.exportAsViewConfString(),
-                currentViewConf = currentViewConfStr && JSON.parse(currentViewConfStr);s               
-                //  currentViewConf && _.extend(nextState, {
-                //      'viewConfig' : currentViewConf
-                //  });
-        }
+    //     if (nextProps.viewConfig !== this.props.viewConfig){
+    //         _.extend(nextState, {
+    //             'originalViewConfig' : null, //object.deepClone(nextProps.viewConfig) // Not currently used.
+    //             'viewConfig'         : nextProps.viewConfig,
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/item-pages/HiGlassViewConfigView.js
```diff
@@ -119,21 +118,21 @@ export class HiGlassViewConfigTabView extends React.PureComponent {
...
             // the URI hash as a consequence of changing tabs --or-- reloading current context due to change in session, etc.
             // Export & save viewConfig from HiGlassComponent internal state to our own to preserve contents.
-            var hgc                 = this.getHiGlassComponent(),
-                currentViewConfStr  = hgc && hgc.api.exportAsViewConfString(),
-                currentViewConf     = currentViewConfStr && JSON.parse(currentViewConfStr);
-
-            currentViewConf && _.extend(nextState, {
-                'viewConfig' : currentViewConf
-            });
+            var hgc = this.getHiGlassComponent(),
```
alexkb0009: _(Not overly important - more for later)_
For most part we can refactor declarations of `var` to `const` (if doesn't get reassigned) or `let` (if does).
(Objects/Arrays whose contents changed but don't get reassigned would fall under `const`).

_(Not overly important, but maybe consider changing if seems simple enough or have time)_
`componentWillReceiveProps` is now deprecated as of React 16 due to async/batched updates and may not always work correctly esp. if many updates having in small amount of time, so if possible we should migrate to `componentDidUpdate(pastProps, pastState)` &/or `static getDerivedStateFromProps(currProps, currState)`

src/encoded/static/components/item-pages/components/TabbedView.js
```diff
@@ -68,7 +68,12 @@ export class TabbedView extends React.PureComponent {
     }
 
     static renderTabPane(tabObj, tabIndex = 0){
-        const { key, tab, placeholder, disabled = false, style = null, content } = tabObj;
+        const { key, tab, placeholder, disabled = false, style = null } = tabObj;
+        let { content } = tabObj;
+        if(this.hiGlassTabIndex >= 0){
```
alexkb0009: 1. `renderTabPane` is a static method where `this` refers to the _function itself_ and not any class instance. `this.hiGlassTabIndex` will always evaluate to `undefined` and condition would always evaluate to false.

2. If it _were_ an instance method, would try to use `memoize` library instead for ease of reasoning. Assigning data to `this` (especially without using `setState`) can get very complex/unwieldy quickly so is nice to avoid if possible. If were an instance method, could perhaps be refactored along these lines:
    ```
    static getHiGlassTabIndex(contents){
        return _.findIndex(contents, function (el) { return el.key === "higlass" });
    }
    
    constructor(props){
        ...
        this.memoized = {
            // Maybe - perhaps unnecessary to memoize since not too unperformant
            getHiGlassTabIndex : memoize(TabbedView.getHiGlassTabIndex)
        }
        ...
    }

    renderTabPane(){ // IF INSTANCE METHOD
        ...
        const higlassTabIndex = this.memoized.getHiGlassTabIndex(contents)
    }
    ```
However this is moot anyway since I guess not needed here and changes in HiGlassViewConfigView.js should be sufficient (haven't tested yet). Other consideration is tab-specific or ItemView-specific logic should ideally not live in TabbedView.js or other 'global'/'reusable' places.

utku-ozturk: Re: 1. `renderTabPane` ...

@alexkb0009  Yes, 100% agree with you that adding hiGlassTabIndex creates a spaghetti code. On the other hand, _.map function's third parameter is context (https://underscorejs.org/#map) and sets this of TabbedView.renderTabPane function.

`'children': _.map(allTabs, TabbedView.renderTabPane, { 'hiGlassTabIndex': hiGlassTabIndex })`

this.hiGlassTabIndex in renderTabPane actually returns an integer. By the way, this truth  
not helps to make the code more readable. We'll remove it asap.
```
if(this.hiGlassTabIndex >= 0){
            content = React.cloneElement(content, { 'hiGlassTabIndex': this.hiGlassTabIndex });
}
```

PR #1149, opened 2019-09-03
Genont mutliont

aschroed:
Changed ontology term schema so a term can belong to multiple ontologies.
Implemented an upgrader for this change.
Added some embeddings for ontology_term to facilitate updates via generate_ontology.
NB. Carl - if you think this is to heavy the new embedding is not strictly necessary.  Without it generate_ontology run on a single ontology takes much longer - however, it is not clear how often a single ontology update will need to occur.

Significant changes to generate_ontology.  This has not been fully tested and some unit tests need to be updated (currently commented out) or written for new functions.

Basically you can either update all ontologies in the db (the default behavior) or specify a single ontology to update.  If specifying a single ontology you can also use the --simple flag to only update those terms that come directly from that ontology.

Running with a single ontology will only update information that comes from that ontology (specified as one of the source ontologies) but may leave outdated synonyms, parents, slim_terms and definitions that came from other source ontologies.


src/encoded/commands/generate_ontology.py
```diff
@@ -441,38 +449,69 @@ def connect2server(env=None, key=None):
...
+        tdefs = term.get('definitions')
+        if not tdefs:
+            continue
+        fromont = [p for p in tdefs.keys() if termid.startswith(p)]
+        if fromont:
+            ont = fromont[0]
+            definition[ont] = tdefs[ont]
+        elif len(tdefs) == 1:
+            definition = tdefs  # we've got a single value to deal with
+        for p, defs in definition.items():
```
sbreiff: from lines 491-495 it looks like `definition` will never have more than 1 item. Is this the intended behavior?

src/encoded/commands/generate_ontology.py
```diff
@@ -441,38 +449,69 @@ def connect2server(env=None, key=None):
...
+            term['definition'] = '{} ({})'.format(' -- '.join(sorted(list(set([d.strip() for d in defs])))), p)
+            found = True
+        if found:
+            del term['definitions']
+            continue
+        # if we get here add all the defs by uniquifying with onts
+        for p, defs in tdefs.items():
+            for d in defs:
+                definition.setdefault(d.strip(), []).append(p)
+        dstring = _format_def_str(defintion)
```
sbreiff: looks like a typo (`defintion` should be `definition`)

src/encoded/commands/generate_ontology.py
aschroed: Nope that looks not quite right.  I'll try to fix this and codacy errors when I have a chance but may not be right away.

src/encoded/commands/generate_ontology.py
aschroed: fixed.

PR #1159, opened 2019-09-13
access key fixes

willronchetti:
- This PR addresses 2 issues and makes note of a 3rd one which will require a more involved fix.
- Issue 1: when creating access keys if your Google account email has capitals it will fail (repro: impersonate me on data, try to create an access key). Solved by lowercasing the email.
- Issue 2: when deploying FF locally, if a user has no access keys, loading their user page will fail with a client error. This is because it tries to access the 'length' field of null. It is not clear why this error does not occur on data, as the same 404 access key error is thrown. Solved by checking if access_keys has type undefined prior to trying to get the length.
- Issue 3: users who do not have any access keys are not able to create any from the user page. Because the above 404 is thrown, even if issue 1 is fixed you still cannot create keys. To get around this you can access Jupyter hub, which will automatically create a key, at which point the user now has an entry in the access key index and issue 1 becomes relevant. This PR does not address this issue at this time.
- This PR still needs to be tested on hotseat. I will update when it is done to verify these fixes have worked.


willronchetti:
Since I can no longer repro issue 3 I presume it was a configuration problem on my end. Thus what should probably happen is the access key inserts should be reverted so it's just my two changes and then we can merge this. Looks like the inserts change caused some tests to fail.

PR #1160, opened 2019-09-14
FontAwesome Upgrade to 5.9.0 & Compilation of `shared-portal-components` into ES5 Modules

alexkb0009:
Upgrade FontAwesome library.

Now SCSS is imported from NPM instead, however we still hold the font files as need to be served. Maybe down the road could load them over into static assets folder from node_modules or something in automated way during buildout.cfg run or `npm run build` or similar.

Many icon instances updated and renamed to v5 versions.
Icons now align more easily, however still some cases where might need to nudge one in some way so it'd look better; in these cases, using Bootstrap "align-*" classNames seem to work well.

`.btn` styling has been updated to have slightly higher padding-top value than Bootstrap default so that the font-family we're using appears more aligned with icons. Text in buttons appears slightly lower than vertical center but in relatively pleasant way and not misaligned.


gulpfile.js
```diff
@@ -48,6 +50,46 @@ function watch(){
...
+    console.log(
+        "`@hms-dbmi-bgm/shared-portal-components` directory is",
+        isLinked ? "sym-linked to `" + sharedComponentPath + "`." : "NOT sym-linked."
+    );
+
+    if (!isLinked){ // Exit
+        done();
+    }
+
+    // Same as shared-portal-components own build method, but with "--watch"
```
alexkb0009: Also uses 4DN's `@babel/cli` (and config) instead of shared-portal-components, but this shouldn't mater as should be equal.

To be safe & future-proof, and not instal babel/cli in 4DN, perhaps could run something like `sharedComponentPath/node_modules/.bin/babl` instead of `npx babel` ... maybe.

PR #1183, opened 2019-10-17
New snovault

carlvitzthum:
Corresponds with [snovault release 1.3.0](https://github.com/4dn-dcic/snovault/releases/tag/v1.3.0)

Main changes:
- dev_servers.py and nginx-dev.conf are moved here from snovault. Slight adaptations and change of `snovault.load_test_data` setting to `load_test_data`.
- `FourfrontRoot.__acl__` is set manually in root.py. Previously it was inherited from snovault `Root`
- Random change: `TrackingItem.STATUS_ACL` changed to include `ALLOW_OWNER_EDIT` for multiple statuses. This is so the tracking items generated by Jupyterhub to track user sessions can be patched using the user's access keys. This is done by setting `submitted_by` to given user on the tracking item


willronchetti:
Needs resubmit

PR #1194, opened 2019-10-31
Biosrc group

aschroed:
- Added a biosource_category calcprop to Biosource - with facet
- Extended it with biosample_category in Biosample
- Embedded biosample_category in ExperimentSet - with facet
- tests for calcprops


src/encoded/types/biosource.py
```diff
@@ -82,36 +82,114 @@ def biosource_name(self, request, biosource_type, individual=None,
...
+                           cell_line=None, cell_line_tier=None, tissue=None,
+                           modifications=None):
+        oterms = self.registry['collections']['OntologyTerm']
+        tid2cat = {
+            'EFO:0003042': 'H1-hESC',
+            'EFO:0002784': 'GM12878',
+            '4DN:0000014': 'HFF (c6 or hTERT)',
+            'EFO:0009318': 'HFF (c6 or hTERT)',
+            '4DN:0000001': 'HFF (c6 or hTERT)',
+            '4DN:0000005': 'WTC-11',
```
sbreiff: Would we want to also classify 4DN:0000267, 4DN:0000266, EFO:0009747 into WTC-11?

src/encoded/schemas/experiment_set.json
```diff
@@ -133,6 +133,9 @@
         "experiments_in_set.biosample.biosource.cell_line_tier": {
             "title": "Sample Tier"
         },
+        "experiments_in_set.biosample.biosample_category": {
+            "title": "Sample Category"
+        },
```
sbreiff: would we want this facet in experiment collection view as well or no?

src/encoded/types/biosource.py
aschroed: Koray said that he was going to change the Biosources linked to the 4DN terms to WTC-11 and we should probably obsolete those 2 terms he created.  And I think we should probably give the EFO term the preferred name of WTC-11 and get rid of our current 4DN term for WTC-11.  We had created that term before we added preferred_name stuff to OntologyTerms I think or else I hadn't realized that GM25256 was the same as WTC-11.

src/encoded/schemas/experiment_set.json
aschroed: We certainly could - do you think it would be useful?  The original motivation was for matrix construction but I guess it would be useful for searching experiments.

sbreiff: I'm thinking about it, between Organism, Sample Type, and Biosource maybe another Biosource-related facet would be overkill. Easier to work with than the Biosource facet though because it has 100+ options. Maybe skip it for now and we can add it back later if we think it will be useful.

PR #1199, opened 2019-11-04
New Cypress Tests Added For Multi/Single Select Search Pages

utku-ozturk:
To increase UI tests coverage, new Cypress test files introduced:

1. Search page multi selection case (`currentAction=multiselect`, file: `08a_search_multiselect.js`)
2. Search page single selection case (`currentAction=selection`, file: `08b_search_single_select.js`)
3. Browse page test is extended by adding tasks to compare filter by file type and download button's count text. (`03d_browse_views_files_selection.js`, see screenshot below.)
4. `03d_browse_views_files_selection.js` test is failed due to a restriction in browse page. (Browse page's select all is permanently disabled if the result count is over a threshold value.) Search result is narrowed by adding a new parameter in search query.

Note that current select page tests (08a/b) are not running as popup window and not really mimics the current behaviour of data portal, since Cypress not supports multiple browser windows as described in https://docs.cypress.io/guides/references/trade-offs.html#Multiple-tabs . Even if 08a/b tests have limited capabilities, it increases the test coverage.

Screen Shot 2019-11-04 at 14 44 07


deploy/post_deploy_testing/cypress/integration/03d_browse_views_files_selection.js
```diff
@@ -1,154 +1,191 @@
...
+    context('Ensure "Select All Files" & related features are functional.', function () {
 
-        before(function(){
-            cy.visit('/browse/?award.project=4DN&experimentset_type=replicate&type=ExperimentSetReplicate');
+        before(function () {
+            cy.visit('/browse/?award.project=4DN&experimentset_type=replicate&type=ExperimentSetReplicate&experiments_in_set.experiment_type.display_title=CUT%26RUN');
         });
 
-        it('We have some counts in QuickInfoBar to compare selected file counts against.', function(){
+        it.skip('We have some counts in QuickInfoBar to compare selected file counts against.', function () {
```
alexkb0009: Maybe remove this `.skip` (if was just left in after testing and forgot to revert)

PR #1200, opened 2019-11-04
Remove sysinfo

willronchetti:
- Removes the sysinfo object
- Configures search to acquire static sections directly


src/encoded/root.py
```diff
@@ -1,5 +1,6 @@
 import json
 import requests
+import datetime
```
carlvitzthum: Unused import?

src/encoded/search.py
```diff
@@ -78,6 +78,16 @@ def search(context, request, search_type=None, return_generator=False, forced_ty
     principals = request.effective_principals
     es = request.registry[ELASTIC_SEARCH]
 
+    # Get static section (if applicable) when searching a single item type
+    if (len(doc_types) == 1) and 'Item' not in doc_types:
+        search_term = 'search-info-header.' + doc_types[0]
+        static_section = request.registry['collections']['StaticSection'].get(search_term)
+        if static_section:
+            item = static_section.model.source['embedded']
```
carlvitzthum: Should use `static_section.properties` here

src/encoded/search.py
```diff
@@ -78,6 +78,16 @@ def search(context, request, search_type=None, return_generator=False, forced_ty
...
+            result['search_header'] = {}
+            result['search_header']['content'] = item['body']
```
carlvitzthum: Should use `item['content']`, which I guess can be different from `body` (confirmed w/ Alex)

PR #1201, opened 2019-11-04
Added override_biosource_name that can be used in biosurce_name calcp…

aschroed:
- override_biosource_name field in biosource schema
- added test
- added EFO term ID for lookup map in biosample

### DIFF
--- a/src/encoded/schemas/biosource.json
+++ b/src/encoded/schemas/biosource.json
@@ -126,10 +126,10 @@
             "linkTo": "Protocol"
         },
         "override_biosource_name": {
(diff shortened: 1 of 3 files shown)


src/encoded/schemas/biosource.json
```diff
@@ -124,6 +124,12 @@
             "type": "string",
             "lookup": 100,
             "linkTo": "Protocol"
+        },
+        "override_biosource_name": {
+            "description": "A override name for the biosource",
```
sbreiff: could add a "title" if it will appear in web UI

PR #1203, opened 2019-11-07
NavBar: NavItems with BigDropdowns

alexkb0009:
## Changes
- Added more NavItems - "Data" + "Resources"
- Added new master-insert for `/resources`, with some inserts moved over from `/help`
- Modularized BigDropdownMenu into multiple parts which can be put together depending on intents.
- Used BigDropdownMenuTree for "Resources" + "Help", custom view for Data dropdown (links to Search URIs, for most part).
- Replaced User account dropdown with BigDropdown (previously React-Bootstrap component was used).
- Cleaned up some extraneous code and moved into `globals.js`:
  - Function for getting Google Analytics tracking ID.
  - `portalConfig` which now only used for `portal_title` anyway.
  - New global function `memoizedUrlParse` to use in place of `url.parse` when parsing current `href` that's passed down from Redux store. `url.parse` is used frequently in many places so made sense to optimize it somewhat. Though not much practical performance improvement is noticed.
- In AccountNav.js (previously UserAccountDropdown), pass in `auth0Options` to LoginController, mirroring CGAP. Necessary as authentication processes starting to diverge (e.g. enable Partners login on CGAP but not on 4DN, use different logos, etc).

### Interesting Bits
`ReactDOM.createPortal` is used to render BigDropdownContainer into `overlaysContainer` (reference to element). This is carried over from existing implementation, though in hindsight a different approach could be used to better effect. However current approach works well enough and offers some other benefits such as simpler re-usability. Each NavItem renders out a different portal, though they all use the same `overlaysContainer` element. By React's design, when 2 portals use the same node element, they will append to it (and not replace all contents).

One of the 'issues' in using React portal(s) is that in the 'process' of changing to a different dropdown (given that one is already open), one component/elemented will become umounted/removed, and then another will get mounted into overlaysContainer. There is no way using React.Portal (AFAIK; had attempted to use key param and couple other ideas) to simply re-render the BigDropdown container with different props/data which would allow for more variety of (& 'smoother') CSS transitions since no new DOM element would get added, just existing DOM element attributes changed.

In this process of switching different view, the unmounting + mounting does not happen in the same browser repaint (likely the 2 React portals both trigger independent repaints, though not researched enough) so there is a "jump" of some kind, for example the BigDropdown, including the half-opaque background, will disappear for a split second when switching to different NavItem.

We used CSS trickery to work around this - namely, once we open a new NavItem and an existing one is open; we delay the closing of the existing BigDropdown for a few milliseconds using `closingDropdownID` state + CSS transitions (e.g. along lines of `dropdown.is-closing { transition: visibility 0s 0.01s; }` & similar) in order to smooth over the split second where it would otherwise disappear and new BigDropdown element has yet to appear. To avoid having 2 semi-opaque background stack and "blend" together to create a slightly-more-opaque background for a split-second (or longer), we apply another "CSS trick" in that we only apply the background-color to the bottom-most (in `z` coordinate, by default, first child of a container is at bottom if no 'z-index' CSS modifiers) via `&:first-child { background-color: rgba(0,0,0,0.7); }`.


src/encoded/static/store.js
```diff
@@ -50,7 +50,7 @@ export const reducers = {
         return state;
     },
 
-    'browseBaseState' : function(state='only_4dn', action){
+    'browseBaseState' : function(state='all', action){
```
alexkb0009: Changes default state of '[Show] External Data' toggle to be true.

alexkb0009: But on UI side only, also default redirection from /browse/ (and other things like, metadata.tsv, barplot aggs (I think..)) changed to exclude `&award.project=4DN` param in search.py.

src/encoded/tests/data/inserts/experiment_set_replicate.json
```diff
@@ -21,7 +21,7 @@
         "static_headers": [
             "item-page-headers.ExperimentSet.data-usage-guidelines",
             "release-updates.TEST1",
-            "help.analysis.file-formats.hic",
+            "resources.analysis.file-formats.hic",
```
alexkb0009: (Re: all insert changes-)
Moved some `Page` & `StaticSection` inserts over to 'Resources' NavItem/Dropdown, by setting as `children` (or children of children) of new insert for page with `name`/url `/resources`.

src/encoded/static/store.js
utku-ozturk: If this PR will not be merged shortly, default External Data = true feature could be merged separately into master.

src/encoded/static/components/navigation/components/AccountNav.js
```diff
@@ -0,0 +1,207 @@
...
+    const { first_name: firstName = "Account", last_name: lastName = null } = userDetails;
+    const introTitle = firstName + (lastName ? " " + lastName : "");
+
+    const viewProfileAction = _.findWhere(userActions, { id: "profile" });
+    const viewProfileURL = (viewProfileAction && viewProfileAction.href) || "/me";
+    const renderedActions = userActions.map(function(action, idx){
+        const { id: actionID, title: actionTitle } = action;
+        const targetHref = getActionURL(action, href);
+        const isActive = isActionActive(action, href);
+        let prepend = null;
```
utku-ozturk: too minor fix: replacing if-if-if statement w/ if-elif-elif could be better, since actionID is const and enter to single if block.

src/encoded/static/components/viz/BarPlot/UIControlsWrapper.js
```diff
@@ -29,8 +30,8 @@ export class UIControlsWrapper extends React.PureComponent {
 
     static contextualView = memoize(function(href){
         if (href){
-            // Hide on homepage.
-            const hrefParts = url.parse(href);
+            // Hide on homepage. I guess could have used PageTitle.isHomePage here instead.
```
Bianca-Morris: Out of curiosity, would there be a difference in functionality between the two?

alexkb0009: Nah... just realizing there's a `PageTitle.isHomePage` func that literally does the same thing. Self note to cleanup some code :-P.

Actually on CGAP we've refactored the PageTitleSection to be much cleaner / customizable that would wanna propagate to 4DN repo and this would likely entail getting rid of `PageTitle.isHomePage` so current approach (if hrefParts.pathname === "/") makes more sense prly.

PR #1209, opened 2019-11-15
Added ACTB_count fields to fastq files schema

Lmercadom:


src/encoded/schemas/experiment_seq.json
```diff
@@ -122,6 +122,13 @@
             "type": "number",
             "lookup": 113,
             "use_for": ["MARGI"]
+        },
+        "rna_strandness":{
```
aschroed: Should be 'strandedness'
suggest that remove the rna_ from the propname and just go with 'strandedness'

src/encoded/schemas/experiment_seq.json
```diff
@@ -122,6 +122,13 @@
...
+            "title": "RNA strand-specific direction",
+            "description": "The direction of the RNA strand used for library preparation",
```
aschroed: title: RNA library strandedness
description: For RNA libraries, do reads correspond to the forward RNA strand, the complementary reverse strand, or both (unstranded)

PR #1228, opened 2019-12-27
Display Search Header Static Section: search-info-header.<Item type>

utku-ozturk:
Pls handle this PR w/ https://github.com/4dn-dcic/shared-portal-components/pull/19

**Problem**: Previously sysinfo item is used to map certain static sections to searches for specific item types. When a search on one item type was done, the given static section content was shown at the top of the search page. Now, search.py will look for a static section with name `search-info-header.<Item type>` when searching for one type, and add it to the search response. At some point, the search JS needs to be adjusted to use the new content in the response instead of looking for the now-deprecated sysinfo object.

**Solution**: Minor updates for both back-end and front-end solved the problem:

Screen Shot 2019-12-28 at 00 25 29

### DIFF
--- a/src/encoded/search.py
+++ b/src/encoded/search.py
@@ -92,6 +92,7 @@ def search(context, request, search_type=None, return_generator=False, forced_ty
             result['search_header'] = {}
             result['search_header']['content'] = item['content']
             result['search_header']['title'] = item.get('title', item['display_title'])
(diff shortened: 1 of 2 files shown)


utku-ozturk:
[search-info-header.<Item type>](https://trello.com/c/ad4wSBjT/39-search-info-headeritem-type)

PR #1230, opened 2020-01-06
New 9a Cypress Biosample Create Test

utku-ozturk:
New test added for the case below:
- Cypress test for Non-admin submitter can create biosample

### DIFF
--- a/deploy/post_deploy_testing/cypress/integration/09a_create_biosample.js
+++ b/deploy/post_deploy_testing/cypress/integration/09a_create_biosample.js
@@ -1,0 +1,90 @@
+
+import { navUserAcctDropdownBtnSelector, navUserAcctLoginBtnSelector } from './../support/variables';
+import { isUUID } from '@hms-dbmi-bgm/shared-portal-components/es/components/util/object';
+
+/**
+* Test you can visit the Biosample create page.
+*/
+describe('Biosample create page', function () {
+
+    context('Cypress test user profile create biosample page', function () { 
+
+        var testItemsToDelete = [];
+        it('Main page visit', function () {
+            cy.visit('/');
+        });
+
+        it('Ensure logged in, visit biosample create page ', function () {
+            //Login CypressTest user
+            cy.login4DN({ 'email': 'u4dntestcypress@gmail.com' }).end()
+                .get(navUserAcctDropdownBtnSelector).then((accountListItem) => {
+                    expect(accountListItem.text()).to.contain('Frontend Cypress');
+                }).end()
+            cy.visit('/search/?type=Biosample').get(".above-results-table-row .btn").should('contain', 'Create New')
+                .get("a.btn.btn-primary.btn").should('contain', 'Create New').click().end().wait(10000)
+            //Submit create biosample data name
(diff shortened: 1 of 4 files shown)


j1z0:


```diff

Complexity increasing per file
==============================
- deploy/post_deploy_testing/cypress/integration/09a_create_biosample.js  1
         

Clones added
============
- deploy/post_deploy_testing/cypress/integration/09a_create_biosample.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4769668&bid=15896468)

Bianca-Morris:
Code looks good; can I ask how I would test this on my local machine? Looks like there are a bunch of different cypress commands in package.json. Which one(s) should I be testing with?

alexkb0009:
There's a user for testing already located in github.com/4dn-dcic/fourfront/blob/master/src/encoded/tests/data/master-inserts/user.json#L134-L145.

It may or may not be worthwhile to just re-use this account. Things in master-inserts folder also gets propagated (or meant to get propagated) to all environments, however things in data/inserts does not (and only goes to localhost). So it's probably less effort to just re-use the existing account rather than then having to go in and POST new user `u4dntestcypress@gmail.com` to every environment (data, hotseat, etc.)

alexkb0009:
@Bianca-Morris 
`cypress:open` Will open the Cypress UI from which can select test(s) to run on local and observe the test as is happening in realtime. cypress-open runs against data.4dnucleome.org.

`cypress:open-<something>` variants do the same, but run against different environments. Likely, `cypress:open-local` would be most relevant to test locally, since test(s) depend on new user in inserts folder which would only ever appear on localhost. Otherwise, I often test/run tests against data.4dnucleome.org.

`cypress:test` commands will run through all the tests against the environment as if are running from Travis and will _not_ present a user interface to view tests as they are ran. `recorded` variants will also upload to Cypress dashboard, assuming $CYPRESS_KEY environment variable is present on local machine.

For all tests to be able to login, requires an Auth0Secret and Auth0Client key on local machine as well. Tests which don't require login should run/pass fine though if not present.

(We might change where all such keys are stored in future; maybe require VPN connection to HMS to get, etc.)

utku-ozturk:
> It may or may not be worthwhile to just re-use this account.

@alexkb0009 new `u4dntestcypress@gmail.com` created since we should do testing a user from an external lab, and `ud4dntest@gmail` is not from an external lab. `u4dntestcypress`'s lab and award is already exist in `data`, we just added it in insert to test in `local`. We post user to `data` manually. You are right, master and hotseat not having `u4dntestcypress@gmail` currently, unfortunately we need to post them too. 
 
Time by time new tests will be introduced, so we may need users for testing in the future
Apparently it doesn't make sense to add new user for each case. If there is a way to reuse already existing ud4dntest@gmail user (PATCHing sounds not good and practical), it would be a better approach?

utku-ozturk:
[Cypress test for Non-admin submitter can create biosample.](https://trello.com/c/k7Qjnaxd/38-cypress-test-for-non-admin-submitter-can-create-biosample)

PR #1234, opened 2020-01-10
Added first line format field to fastqc schema

Lmercadom:


src/encoded/schemas/file_fastq.json
```diff
@@ -154,6 +154,14 @@
             "exclude_from": ["submit4dn", "FFedit-create"],
             "permission": "import_items",
             "lookup": 1001
+        },
+        "first_line_format":{
+            "title": "First line of the fastq file",
+            "description": "This is how the first line of the fastq file looks like",
```
aschroed: What is the actual value that you want in this field?  Is it the actual first line of the file that this metadata is for or is it some description of the fields included in the line?  You can include an example in the description if that would be clearer.  If it's the actual first line then why not call the field that 'file_first_line'.  And the description would be 'First line of the fastq file'.

Lmercadom: Yes, it is the actual first line. All of it: "@K00124:595:HGJ5YBBXX:7:1101:1822:2617 2:N:0:GAGTTAGC"

PR #1237, opened 2020-01-16
Deploy Change

willronchetti:
- Change `creating_mapping_on_deploy` to allow the caller to configure whether or not to trigger a full re-index
- By default do not trigger a full re-index unless `create_mapping_on_deploy` is passed the `--wipe-es` option explicitly.

### DIFF
--- a/src/encoded/commands/create_mapping_on_deploy.py
+++ b/src/encoded/commands/create_mapping_on_deploy.py
@@ -131,6 +131,7 @@ def main():
     )
     parser.add_argument('config_uri', help="path to configfile")
     parser.add_argument('--app-name', help="Pyramid app name in configfile")
+    parser.add_argument('--wipe-es', help="If specified on webprod will wipe ES", action='store_true', default=False)
 
     args = parser.parse_args()
     app = get_app(args.config_uri, args.app_name)
@@ -144,8 +145,13 @@ def main():
         env = app.registry.settings.get('env.name')
         if 'webprod' in env:
             if data_env != env:
-                log.info("looks like we are on staging, run create mapping without check first")
-                run_create_mapping(app, check_first=False, item_order=ITEM_INDEX_ORDER)
+                log.info("Looks like we are deploying staging, checking --wipe-es")
+                if args.wipe_es:
+                    log.info('--wipe-es specified - obliging')
+                    run_create_mapping(app, check_first=False, item_order=ITEM_INDEX_ORDER)
+                else:
+                    log.info('--wipe-es not specified - will not wipe ES')
+                    run_create_mapping(app, check_first=True, item_order=ITEM_INDEX_ORDER)
                 return
         # handle mastertest ... by blowing away all data first
         if 'mastertest' in env:


j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/commands/create_mapping_on_deploy.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4834381&bid=16057364)

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -144,8 +145,13 @@ def main():
         env = app.registry.settings.get('env.name')
         if 'webprod' in env:
             if data_env != env:
-                log.info("looks like we are on staging, run create mapping without check first")
-                run_create_mapping(app, check_first=False, item_order=ITEM_INDEX_ORDER)
+                log.info("Looks like we are deploying staging, checking --wipe-es")
```
netsettler: Hmm. In my view, this logic about how we tell what we're deploying to needs to be cleaned up to be less ad hoc in nature, but that wouldn't be a necessarily local change. The minimal thing would be to isolate the `data_env != env` logic in a separate function that knows what the accepted technique is, just in case other things need to ask this question so they all use the same mechanism. A better strategy would be to declare the intent that this be staging and have the choice to separate that data env drive off that declaration and poll that same piece of declarative information here. But for now the technique here looks good enough and uses a reasonable heuristic.

I might prefer the message say something like `"Looks like we are not using the production data store,..."` rather than `"Looks like we are deploying staging,..."` so it isn't saying something that is stronger than what it knows. It also might be ok to say "Looks like this is a staging deploy,..." (note change to "a staging deploy" rather than "deploying staging", which would suggest there were different deploys that count, rather than it has verified a domain name or something. But this is a smallish concern and need not hold anything up now. I just find that sometimes if something goes wrong, people read these messages and think that something has been checked that hasn't been, and it can throw them off the track of what they should be diving further into.

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -144,8 +145,13 @@ def main():
...
+                if args.wipe_es:
+                    log.info('--wipe-es specified - obliging')
+                    run_create_mapping(app, check_first=False, item_order=ITEM_INDEX_ORDER)
+                else:
+                    log.info('--wipe-es not specified - will not wipe ES')
+                    run_create_mapping(app, check_first=True, item_order=ITEM_INDEX_ORDER)
                 return
         # handle mastertest ... by blowing away all data first
         if 'mastertest' in env:
```
netsettler: Should this be an `elif`?  Assuming I'm not mis-reading here, the comment is worrisome in context of the above, and it seems possible a weird env name like `webprod-mastertest` could cause havoc, running everything twice, but in any case if the arms are disjoint normally anyway it seems worth making that clear.

netsettler: There's a call to log.info that says  "looks like we are NOT on staging or mastertest " that also looks like it belongs in some elif, too, and I'm less certain it's properly enumerating the cases there. This is messiness I do NOT expect to see cleaned up on this PR, as it feels like it would drift too far from cause. Even the `elif` question might be too much expansiveness. But I don't like this kind of control-flow vagueness of text message and haphazardness of expression on something so critical as wiping data.

PR #1239, opened 2020-01-23
Attempt to fix wsgi appdeploy problem

netsettler:
Trying to fix a bug in what Carl left us, this creates an empty file in place of the one that is wants to get deleted but that still tries to run. In particular, the eb code seems to get a list of files it plans to run before we delete them, and so then tries to run a deleted file.

I'm also piggy-backing a few other changes in this PR to .gitignore and the doc.

### DIFF
--- a/.ebextensions/05_set_wsgi.config
+++ b/.ebextensions/05_set_wsgi.config
@@ -24,10 +24,6 @@ files:
       for old_wsgi_file in old_wsgi_hooks_conf:
           if os.path.exists(old_wsgi_file):
               os.remove(old_wsgi_file)
-              if os.path.endswith(".py"):
-                  # Write .py files anew as an empty file instead.
-                  with open(old_wsgi_file, 'w'):
-                      pass
 
       MY_APACHE_TEMPLATE = '''
       # Customized wsgi.conf.  If you're seeing this, good!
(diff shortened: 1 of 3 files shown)


netsettler:
It seems that we have an alternate workaround, to just get a fresh set of files, but I think we should retain this change because it's definitely going to fail if those files ever exist and this might ameliorate that if things were to get set wrong in the future. (If we really expect the files to not be there the right fix is to remove the loop that deletes them.)

willronchetti:
I tried to deploy this to mastertest and got: `Command failed on instance. Return code: 1 Output: container_command 081_elastic_search_mapping in .ebextensions/20_packages.config failed.`

netsettler:
> I tried to deploy this to mastertest and got: `Command failed on instance. Return code: 1 Output: container_command 081_elastic_search_mapping in .ebextensions/20_packages.config failed.`

Wow, it's hard to see how this could have caused that. Maybe there are other problems with the show-through that it gets to because this succeeds.  I'll take my change out of this PR after all, and then merge it with only the readme and gitignore changes, that way we'll know this isn't making new headaches we don't need.

willronchetti:
Feel free to just merge this now that it's just semantic changes

PR #1242, opened 2020-01-27
Misc test fixes (2020-01-26)

netsettler:
* Some cosmetic and substantive changes to README.rst

  - Better section headings.
  - Better choice of when something is an RST code block. (PyCharm really helped to debug this.)
  - Add substantive notes about how to repair various kinds of problems such a JDK issues, zlib issues, and various brew issues.
  - Be more careful about suggesting `brew upgrade`.

* Fix an import in test_create_mapping.py since it was failing locally for Kent (maybe a Python 3.4 vs 3.5 issue?)

* In test_types_init_collections, fix some uses of `datetime.datetime.now` that needed to be `datetime.datetime.utcnow`. At the same time, use more abstract operations to extract the date, not string operations, and factor out some imports to the top of the file.

### DIFF
--- a/src/encoded/tests/test_types_init_collections.py
+++ b/src/encoded/tests/test_types_init_collections.py
@@ -229,12 +229,10 @@ def test_tracking_item_display_title_google_analytic(google_analytics):
 
 
 def test_tracking_item_display_title_download(download_tracking):
-    from datetime import datetime
     assert download_tracking.get('display_title') == 'Download Tracking Item from ' + utc_today_str()
 
 
 def test_tracking_item_display_title_other(jupyterhub_session):
-    from datetime import datetime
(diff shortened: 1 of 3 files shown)


src/encoded/tests/test_types_init_collections.py
```diff
@@ -1,7 +1,12 @@
+import datetime
 import pytest
 pytestmark = [pytest.mark.setone, pytest.mark.working, pytest.mark.schema]
 
 
+def utc_today_str():
```
willronchetti: I suspect we want this in other places as well, might be worth pulling into a different file

netsettler: Yeah, I was thinking it should be in another repo that is shared. Utils or the like.

PR #1248, opened 2020-01-31
Added note field to file schema

Lmercadom:


src/encoded/types/file.py
```diff
@@ -730,6 +730,19 @@ def create(cls, registry, uuid, properties, sheets=None):
...
+        "description": "Notes that go into the metadata.tsv file",
+        "type": "string"
+    })
+    def tsv_notes(self, request, notes_to_tsv=None):
+        if notes_to_tsv is None:
+            return ''
+        elif len(notes_to_tsv) == 1:
+            notes_to_tsv_string = notes_to_tsv[0]
+        else:
+            notes_to_tsv_string = ','.join(notes_to_tsv)
```
sbreiff: Feel free to leave it this way if you want, but fyi if you use `','.join(lst)` on a list with one element, it will return that element as a string without the comma. So, if you wanted, you could remove the `elif` clause.

Lmercadom: Thank you for the tip Sarah, I will change it so the code is cleaner

PR #1254, opened 2020-02-11
added individual relations

andreacosolo:
- unidirectional relations (from F1 to parental strain)
- all individuals, excluding humans
- related individual has to be from the same species


j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/types/individual.py  15
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/types/individual.py  16
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  2
- src/encoded/types/individual.py  16
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

src/encoded/types/individual.py
```diff
@@ -118,3 +134,121 @@ class IndividualZebrafish(Individual):
...
+    relations_counter = {}
+    relations_unique = {}
+    for a_related_individual in related_individuals:
+        parent = a_related_individual.get('individual')
+        print(parent)
+        parent_props = get_item_if_you_can(request, parent, 'individuals', frame='raw')
+        print(parent_props)
+        if parent_props:
+            parent_organism = parent_props.get('organism')
+            parent_uuid = parent_props.get('uuid')
```
sbreiff: This parent_uuid doesn't get created if parent_props aren't found, which I think is causing some problems. You can either indent all the code below another level or add an else clause.

Don't forget to take out the frame=raw a few lines above, as well.

andreacosolo: Does it make sense to keep the if clause at all? 
Given a parent individual, the get_item_if_you_can() will always return the item, unless something is very wrong. Or would this be a way to test if the wrong item has been added as parent (e.g. enzyme, instead of individual)?

src/encoded/tests/test_types_individual.py
```diff
@@ -0,0 +1,104 @@
...
+    errors = res.json['errors']
+    assert errors[0]['name'] == 'Individual relation: different species'
+
+
+def test_validate_individual_relation_valid_patch(testapp, award, lab, mouse_child, mouse_individual):
+    res = testapp.post_json('/individual_mouse', mouse_child, status=201)
+    assert not res.json.get('errors')
+    patch_body = [{'relationship_type': 'derived from', 'individual': mouse_individual['@id']}]
+    res2 = testapp.patch_json(res.json['@graph'][0]['@id'], {'individual_relation': patch_body})
+    assert not res2.json.get('errors')
```
sbreiff: If you wanted, you could add/edit lines here so that a mother and father parent strain are patched. This would ensure that the item passes validation when there are 2 parents and not just one.

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  3
- src/encoded/types/individual.py  14
         

Clones added
============
- src/encoded/tests/test_types_individual.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

src/encoded/tests/test_types_individual.py
andreacosolo: Good idea. Done.

src/encoded/types/individual.py
```diff
@@ -118,3 +137,114 @@ class IndividualZebrafish(Individual):
...
+    limited to two relations per individual (max 1 paternal and 1 maternal),
+    not self relations, and unique (no duplicate relations).
+    '''
+    data = request.json
+    related_individuals = data.get('individual_relation')  # a list of dicts
+    if related_individuals is None:
+        return
+    any_failures = False
+    # check if the individual info is in request (POST) or in context (PATCH)
+    try:
```
sbreiff: I was having issues deploying local and loading the page when I pulled your branch. The following suggestion (replacing the whole try/except block) fixed the issue for me, though rather than just accepting it you might want to try yourself and see if Andy has any input since he might have more experience with this part of the code than me. 

```suggestion
    individual_uuid = data.get('uuid')
    organism_id = data.get('organism')
    # try:
    #     individual_uuid = data['uuid']
    #     organism_id = data.get('organism')
    # except KeyError:
    #     individual_uuid = str(context.uuid)
    #     individual = get_item_if_you_can(request, individual_uuid, 'individuals')
    #     organism_id = individual.get('organism')
```

sbreiff: Further note: organism is a required property for individual, so it should always be present and I don't think it would ever make sense to patch a different organism. uuid should also always be in the post, unless I'm misunderstanding the process.

andreacosolo: I could deploy local without issues, so I am not sure what the issue might be.

The whole try/except block is necessary because otherwise the patch test fails. When you patch just the individual_relations (as in the test), the information about the individual being patched (e.g. its uuid) is not in the variable request, but in context. This block accounts for that.

sbreiff: Alright, I'll take your word for it then and look into why I can't deploy.

src/encoded/schemas/individual.json
```diff
@@ -54,6 +54,37 @@
                 "mixed"
             ]
         },
+        "individual_relation": {
+            "title": "Individual Relations",
+            "description": "Related individuals",
+            "type": "array",
```
aschroed: You can add here:
"uniqueItems": true,

It doesn't change the need for the check in the validator that looks for different relationship types for the same individual but it will prevent mistakes in submitting multiple of the same individual_relations objects.

src/encoded/types/individual.py
```diff
@@ -118,3 +137,114 @@ class IndividualZebrafish(Individual):
...
+    '''Validates that individual relations are within the same species,
+    limited to two relations per individual (max 1 paternal and 1 maternal),
+    not self relations, and unique (no duplicate relations).
+    '''
+    data = request.json
+    related_individuals = data.get('individual_relation')  # a list of dicts
+    if related_individuals is None:
+        return
+    any_failures = False
+    # check if the individual info is in request (POST) or in context (PATCH)
```
aschroed: In order to save a call to get_item_if_you_can I might change lines 154 to 160 to something like:

    if 'uuid' in data:  # it's a POST
        individual_uuid = data['uuid']
    else:  # get it from the context it's a PATCH
        try:
            individual_uuid = str(context.uuid)
        except Exception:  # would be very wonky if this happened
            pass
        finally:
            data = context.properties  # organism will be here now
    organism_id = data.get('organism')

aschroed: This change is probably optional but another possible better practice is that the try clause in my example includes only a signal statement to try rather than the 2 in your version either of which could potentially cause the exception.  Not sure.

src/encoded/schemas/individual.json
andreacosolo: OK!
And is there a way in the schema to limit the number of individual_relation items to max 2?

aschroed: You can try:

"maxItems": 2,

But that might have either been introduced in a later version of json schema than we are using or require some code someplace in snovault to do that type of validation.  Worth a try though.

sbreiff:
My final comment is that once this is in production, Submit4DN will have to be edited so that individual_relation is added to the list of second round patches. Alternatively, it might be possible to edit code in import_data so that instead of a hard-coded list, we get all the profiles and find all the fields that say `"ff_flag": "second round"`.

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  3
- src/encoded/types/individual.py  15
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

aschroed:
I just made a change to the validator to deal with the issue that was occurring when using Submit4DN.  It just switched around a bit how it is determined if it is a post or patch and where it gets associated data from 'context' or 'request'.  I wrote a test that covers that case.

The issue was that Submit4DN includes the uuid in the patch json for second round patches but does not include the organism info automatically.

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  3
- src/encoded/types/individual.py  14
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  3
- src/encoded/types/individual.py  14
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

j1z0:


```diff

Complexity increasing per file
==============================
- src/encoded/tests/test_types_individual.py  3
- src/encoded/types/individual.py  15
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=4985365&bid=16461759)

PR #1260, opened 2020-02-20
Update description for lab title

andreacosolo:


src/encoded/schemas/lab.json
```diff
@@ -39,7 +39,7 @@
         },
         "title": {
             "title": "Title",
-            "description": "A unique name for lab & affiliation use PI's first & last name, Lab, comma, and institute label. (e.g. John Doe Lab, UNI).",
+            "description": "A unique name for lab & affiliation use PI's first & last name, comma, and institute label. (e.g. John Doe, UNI).",
```
netsettler: Is there anything that enforces this, or is this a suggestion? If not just a suggestion, is there a unit test? I looked but could not find any. I might not have looked in the right place. (If one wanted to say "Park Lab, HMS"  or even just "My Famous Freestanding Lab" would something break? What happens if the first and last name of two PI's at the same institute are identical?)

I don't know enough about the conventions in play here to approve this or not, but I figured I'd offer those questions.

andreacosolo: Thanks for your review. Currently there are no tests to enforce it, and I think it is not necessary since it's mostly (if not always) data wranglers who fill out this. As for the possible non-unique cases, the likelihood of such scenario verifying is so low (given the size of the scientific community in the field) that I think it is fine to come up with ad-hoc solutions once it happens.

PR #1266, opened 2020-02-28
Adding embeded views

Lmercadom:


src/encoded/types/experiment.py
```diff
@@ -142,6 +144,12 @@ class Experiment(Item):
         "files.badges.badge.description",
         "files.badges.badge.badge_icon",
         "files.badges.messages",
```
sbreiff: While you're editing the embeds here, you might as well comment out all the files.badges ones for now since we don't have badges for files currently.

src/encoded/types/experiment.py
```diff
@@ -142,6 +144,12 @@ class Experiment(Item):
...
+        "files.quality_metric.display_title",
```
sbreiff: you could delete this line ^^ because display_title always gets embedded when other properties are embedded.

src/encoded/types/experiment.py
```diff
@@ -150,6 +158,27 @@ class Experiment(Item):
...
+        "processed_files.file_type_detailed",
+        "processed_files.quality_metric.url",
+        "processed_files.quality_metric.overall_quality_status",
+        "processed_files.quality_metric_summary.*",
+
+        "other_processed_files.files.href",
+        "other_processed_files.title",
+        "other_processed_files.description",
+        "other_processed_files.type",
+        "other_processed_files.files.accession",
```
sbreiff: accession could probably also be deleted as i think its automatically embedded

PR #1271, opened 2020-03-20
AJAX status events in & display alert messages

alexkb0009:
Should look like:

(See code; relatively small edit)

Maybe could add something like this to lambda response following re: [this StackOverflow discussion/answer](https://stackoverflow.com/questions/32500073/request-header-field-access-control-allow-headers-is-not-allowed-by-itself-in-pr#answer-37228330) also:

```python
response.setHeader("Access-Control-Allow-Methods", "GET,HEAD,OPTIONS"); # idk if anything other than GET needed.
response.setHeader(
    "Access-Control-Allow-Headers",
    "Origin, Accept, X-Requested-With, Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers, Cache-Control, Authorization"
);
```

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -215,6 +215,7 @@ export default class App extends React.PureComponent {
      */
     componentDidMount() {
         const { href, context } = this.props;
+        const { statusMessagesURL = null } = portalConfig;
 
         // The href prop we have was from serverside. It would not have a hash in it, and might be shortened.
         // Here we grab full-length href from window and then update props.href (via Redux), if it is different.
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/app.js
```diff
@@ -297,6 +298,27 @@ export default class App extends React.PureComponent {
...
+                    });
+                },
+                "GET",
+                null,
+                function(err){
+                    console.error("Error getting status events", err);
+                },
+                // Clear headers for CORS
+                {}, //{ "Cache-Control": "max-age=120" }
+                ["Accept", "Content-Type", "Authorization", "X-Requested-With"]
```
netsettler: At @alexkb0009's suggestion, I added more headers that may fix the need for some of this. This is from the current source on the lambda function (not yet in source control, just on AWS's console). These headers are added to all responses from that endpoint (the one you call `statusMessagesURL` here)
```
CORS_HEADERS = {
    # It may be that only GET is needed, but just in case. -kmp&akb 20-Mar-2020
    "Access-Control-Allow-Methods": "GET,HEAD,OPTIONS",
    "Access-Control-Allow-Headers": ",".join([
        "Origin",
        "Accept",
        "X-Requested-With",
        "Content-Type",
        "Access-Control-Request-Method",
        "Access-Control-Request-Headers",
        "Cache-Control",
        "Authorization",
    ]),
    "Access-Control-Allow-Origin": "*",
}
```

src/encoded/static/components/item-pages/WorkflowRunTracingView.js
```diff
@@ -212,12 +212,6 @@ export class FileViewGraphSection extends WorkflowGraphSection {
         };
     }
 
-    componentWillReceiveProps(nextProps){
-        if (this.props.steps !== nextProps.steps){
-            this.setState(checkIfIndirectOrReferenceNodesExist(nextProps.steps));
-        }
-    }
-
```
alexkb0009, line 220: Forgot why this edit was made, maybe will revert until can explore further.

alexkb0009: Ok yeah this is safe to remove. It would return `anyReferenceFileNodes` & `anyReferenceFileNodes` which we don't want in state any longer.

PR #1281, opened 2020-04-01
Partial fix for C4-105, addressing types_workflow.py

netsettler:
This PR addresses one of several problems that `dcicutils.env_utils 0.13.0` can help to fix. I'm pushing out the fixes as they become available.  It's still testing as I write this, but I expect it to pass.

(This will _probably_ also make this work for CGAP, not that this code is used in CGAP, but it's good for us to be able to copy&paste things between these repos where it makes sense.)

I added some unit tests.

### DIFF
--- a/src/encoded/types/workflow.py
+++ b/src/encoded/types/workflow.py
@@ -934,7 +934,7 @@ def pseudo_run(context, request):
     # for testing
     if not env:
         env = 'fourfront-webdev'
-    if env == 'fourfront-webprod2':
+    if env in ['data', 'staging', 'fourfront-green', 'fourfront-blue', 'fourfront-webprod2']:
         input_json['output_bucket'] = 'elasticbeanstalk-fourfront-webprod-wfoutput'
     else:
         input_json['output_bucket'] = 'elasticbeanstalk-%s-wfoutput' % env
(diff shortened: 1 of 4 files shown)


src/encoded/types/workflow.py
```diff
@@ -1,34 +1,23 @@
...
+from collections import OrderedDict, deque
+from dcicutils.env_utils import FF_ENV_WEBDEV, is_stg_or_prd_env, prod_bucket_env
+from inspect import signature
+from pyramid.httpexceptions import HTTPUnprocessableEntity
+from pyramid.response import Response
+from pyramid.view import view_config
+from snovault import calculated_property, collection, load_schema, CONNECTION, TYPES
+from snovault.util import debug_log
 from time import sleep
+from .base import Item, lab_award_attribution_embed_list
```
netsettler, line 20: Mostly these imports are just rearranged (alphabetized) to make it easier to see what's there. The only difference of substance is the introduction of the `dcicutils.env_utils` line.

netsettler:
> Looks good to me. What's yellow?

It's a decoy, just an environment that doesn't exist to detect whether there is different behavior for environments we have not defined. :)

netsettler:
Travis shows this test failed, but I restarted it and it succeeded. It's not decaching that fact, but at this time all looks good. I'll merge this.

PR #1287, opened 2020-04-13
Nested Compatibility

willronchetti:
- NOTE: This PR is a work in progress and serves as documentation for what is needed to enable ElasticSearch mapping with `nested`. This PR will be revisited at a later time, as one of our API's (`bar_plot_chart`) becomes extremely complicated when nested types are introduced. This point is elaborated on below so when picked back up one can jump back in.
- Previously, `bar_plot_chart` was very simple - all fields are mapped with standard types, so for both the metrics and X,Y axis aggregations, there are no nested paths to consider.
- When `nested` is introduced, a lot of simple mechanisms we took advantage of no longer apply for this API.
- Problem 1 - there are now 4 cases to consider for the chart (assuming we are only handling to a depth of 2, which is _essentially_ required in the presence of nested as below will illustrate). The primary aggregation is always on the X axis, so for simplicity I will refer to the primary aggregation as the X aggregation and the 'secondary aggregation' as the Y aggregation (X,Y). The cases are: (nested, nested), (nested, non-nested), (non-nested, nested), (non-nested, non-nested). Note that the symmetric case is in fact handled differently. The common case now becomes (nested, nested).
- Problem 2 - the metrics we compute are all on nested fields and they do not all share the same nested path. In Lucene, the nested path is inherited from the parent aggregations. So these metrics work like normal at the top level but will break on the inner levels depending on which case from problem 1 you are in. For example, if your x-axis is experiment_type, as is the default case, then you inherit `nested_path=embedded.experiments_in_set` in all sub-aggregations. The way around this is to use `reverse_nested`, but when doing so you must once again create a sub-aggregation. 
- Problem 3 - Note that in the case of (nested, nested), your inner aggregation will also be broken if the two axis fields do not share the same nested path, which in principle they could. Note additionally that the same problem could occur in any case that involves nested if the path does not match the metrics (guaranteed to be the case).
- Problem 4 - formatting this for the front-end becomes more complicated as well, as each nested thing requires a sub-aggregation. Previously the buckets were in a well defined format - the metrics themselves will look different depending on which of the 4 cases from problem 1 are present in addition to being different altogether due to nested sub-aggregation. The crux of the issue is - how do you in general extract the bucket counts given the 4 different cases above?
- All 4 of these problems need to be solved before any of this will work. I think a reasonable plan of attack would be devise query strategies for all 4 scenarios and handle each one as they come. It is likely a `AggBuilder` of some sort will need to be written to encapsulate a lot of this complexity.


j1z0:


```diff

Issues
======
+ Solved 9
- Added 6
           

Complexity increasing per file
==============================
- src/encoded/search_utils.py  8
         

Complexity decreasing per file
==============================
+ src/encoded/search.py  -14
         

Clones added
============
- src/encoded/search_utils.py  1
- src/encoded/search.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270&bid=17520728)

src/encoded/visualization.py
```diff
@@ -211,71 +180,97 @@ def bar_plot_chart(context, request):
...
+        if 'embedded' not in field:
+            path = find_nested_path('embedded.' + field, item_type_es_mapping)
+        else:
+            path = find_nested_path(field, item_type_es_mapping)
+        if path:
+            nested_paths[field] = path
+
+    # establish elasticsearch_dsl class that will perform the search
+    search = Search(using=es, index=es_index)
+    search, string_query = build_query(search, prepared_terms, source_fields)
```
j1z0, line 203: Codacy Issue found: [Unused variable 'string_query'](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

src/encoded/search.py
```diff
@@ -819,43 +868,310 @@ def set_filters(request, search, result, principals, doc_types):
...
+                            sub_query = handle_should_query(field, options)
+                            _q[NESTED][QUERY][BOOL][key].append(sub_query)
+
+                        # if we don't have options, our original 'query' is what we need
+                        else:
+                            _q[NESTED][QUERY][BOOL][key].append(query[BOOL][key][0])
+
+                        found = True  # break is not sufficient, see below
+                        break
+                    except:       # Why? We found a 'range' nested query and must add this one separately
```
j1z0, line 1046: Codacy Issue found: [No exception type(s) specified](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

src/encoded/search.py
```diff
@@ -819,43 +868,310 @@ def set_filters(request, search, result, principals, doc_types):
...
+                        continue  # This behavior is absurd. Somehow it knows to combine separate nested range
+                                  # queries with AND, but of course not regular queries and of course you cannot
+                                  # combine the range query here due to syntax  - Will
+        if not found:
+
+            # It's possible we're looking at a sub-query that's wrapped in a (length 1) list
+            if type(query) == list:
```
j1z0, line 1053: Codacy Issue found: [Using type() instead of isinstance() for a typecheck.](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

src/encoded/visualization.py
```diff
@@ -1,8 +1,8 @@
-from pyramid.response import Response
 from pyramid.view import view_config
 from pyramid.httpexceptions import HTTPBadRequest
-from snovault import CONNECTION
+from snovault import CONNECTION, TYPES
```
j1z0, line 3: Codacy Issue found: [Unused TYPES imported from snovault](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

src/encoded/search.py
```diff
@@ -1156,33 +1491,35 @@ def set_facets(search, facets, search_filters, string_query, request, doc_types,
...
-                            'field' : query_field
+                AGGS: {
+                    'primary_agg': {
+                        'stats': {
+                            'field': query_field
                         }
                     }
                 },
-                'filter': {'bool': facet_filters}
+                FILTER: {BOOL: facet_filters}
```
j1z0, line 1501: Codacy Issue found: [Using variable 'facet_filters' before assignment](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

src/encoded/search_utils.py
```diff
@@ -0,0 +1,255 @@
...
+    err_exp = None
+    try:
+        es_results = search.execute().to_dict()
+    except ConnectionTimeout as exc:
+        err_exp = 'The search failed due to a timeout. Please try a different query.'
+    except RequestError as exc:
+        # try to get a specific error message. May fail in some cases
+        try:
+            err_detail = str(exc.info['error']['root_cause'][0]['reason'])
+        except:
```
j1z0, line 241: Codacy Issue found: [No exception type(s) specified](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5350270)

willronchetti:
There is a bug in this when constructing negative queries - they are already wrapped in `must_not` so they should use the `must` key in the sub-query.

PR #1290, opened 2020-04-16
C4-128 set mirror_health setting in FF init

willronchetti:
- `mirror_health` was not set in the FF initializer, which breaks deletion mirroring

### DIFF
--- a/src/encoded/__init__.py
+++ b/src/encoded/__init__.py
@@ -15,6 +15,7 @@ import sys
 from dcicutils.beanstalk_utils import source_beanstalk_env_vars
 from dcicutils.log_utils import set_logging
 from dcicutils.env_utils import get_mirror_env_from_context
+from dcicutils.ff_utils import get_health_page
 # from pyramid.authorization import ACLAuthorizationPolicy
 from pyramid.config import Configurator
 # from pyramid.path import AssetResolver, caller_package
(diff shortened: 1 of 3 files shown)


src/encoded/__init__.py
```diff
@@ -144,8 +145,13 @@ def main(global_config, **local_config):
     # set google reCAPTCHA keys
     settings['g.recaptcha.key'] = os.environ.get('reCaptchaKey')
     settings['g.recaptcha.secret'] = os.environ.get('reCaptchaSecret')
+
     # set mirrored Elasticsearch location (for staging and production servers)
-    settings['mirror.env.name'] = get_mirror_env_from_context(settings)
+    mirror = get_mirror_env_from_context(settings)
+    if mirror is not None:
+        settings['mirror.env.name'] = mirror
+        settings['mirror_health'] = get_health_page(ff_env=mirror)
```
netsettler, line 153: At some point we'll need another name for this than `ff_env` because I think this will actually work for cgap, too, but I understand why that name is in there for now and adjustments would be complex. So this is fine and looks way more robust.

willronchetti: Yeah, that type of change would require a major version update for dcicutils as that keyword argument is littered throughout our codebase.

PR #1300, opened 2020-04-28
Minimize differences between CGAP/FF src/encoded/types/workflow.py

netsettler:
This makes some changes that minimize differences between `src/encoded/types/workflow.py` in Fourfront and in CGAP. This is a change that was a byproduct of some other work done weeks ago. This is just cleaning up a lingering bit of work I didn't finish at the time.

[Beyond Compare](https://www.scootersoftware.com/features.php) is a little better at comparing the proposed changes than `git` is, though now that I've submitted this PR, I see github actually doesn't do a horrible job.

It might actually not be needed, but I'm attaching two such comparisons (bundled into a zip file, [Comparisons.zip](https://github.com/4dn-dcic/fourfront/files/4544300/Comparisons.zip)). In each, the proposed change is in column 1. In the one called Report.html, it's comparing this change to Fourfront master. In the one called Report-vs-cgap.html, it's comparing this formatting of things to cgap-portal, minimizing the differences so that one can see what is actually different between cgap-portal's version of this file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "encoded"
-version = "1.2.7"
+version = "1.2.8"
 description = "4DN-DCIC Fourfront"
 authors = ["4DN-DCIC Team <support@4dnucleome.org>"]
 license = "MIT"
--- a/src/encoded/types/workflow.py
+++ b/src/encoded/types/workflow.py
@@ -2,8 +2,8 @@
 """
 
 import copy
-import cProfile
 import boto3
+import cProfile
 import io
 import json
 import pstats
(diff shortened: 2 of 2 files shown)


netsettler:
@SooLee, if you could, please just look this over and tell me if it raises any red flags for you. It's not intended to be a functional change. I just wanted to minimize the accidental syntactic differences between CGAP and Fourfront on this file so it was easier to see the few ways in which they had intentionally diverged.  

The enclosed comparison bundle contains an HTML file that shows those diffs (though one view will appear way off screen to the right unless you have a very big screen, so you have to scroll right to see the other half of the side-by-side).

I put the Fourfront side of the diff on the left so that you can see the stuff that you're reviewing without scrolling, and only scroll to the right if you wonder what has changed. Also, Beyond Compare uses "blue" to mean "unimportant changes" and "red" to mean "important changes", so the blue parts are things it's saying differ mainly in indentation level because I adjusted some of that.

Let me know if you have questions, either here or on slack, as appropriate.

PR #1310, opened 2020-05-02
Test npm ci

alexkb0009:
After few days if assert from new `git status` line, in Travis tests that `package-lock.json` does not get updated:

- Ensure higlass-multivec _release_ working with `npm ci` (possibly errored because from git branch and not a tag/release).
- Delete git status line
- Merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,10 @@ aws-ip-ranges:
 	curl -o aws-ip-ranges.json https://ip-ranges.amazonaws.com/ip-ranges.json
 
 npm-setup:  # runs all front-end setup
-	npm install
+	npm ci
 	npm run build | grep -v "node_modules\|\[built\]"
 	npm run build-scss
+        git status
 
 moto-setup:  # optional moto setup that must be done separately
 	pip install "moto[server]==1.3.7"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "encoded"
-version = "1.3.3"
+version = "1.3.3abcd"
```
netsettler: This is a bug fix that hopefully won't change the perceived action of the system.
I would call it 1.3.4.

Makefile
```diff
@@ -11,9 +11,10 @@ aws-ip-ranges:
 	curl -o aws-ip-ranges.json https://ip-ranges.amazonaws.com/ip-ranges.json
 
 npm-setup:  # runs all front-end setup
-	npm install
+	npm ci
 	npm run build | grep -v "node_modules\|\[built\]"
 	npm run build-scss
+	git status
```
netsettler, line 17: `git status` might show changes we don't care about, though maybe we could arrange to checkpoint its output before and compare it after to see if something happened anew. That's likely to be a pain to do, so I'd just leave it out for now.

netsettler:
I've added an approval here because I am basically in agreement with this change. I'm assuming the higlass issue is being adequately reviewed by others

PR #1318, opened 2020-05-15
Schema for TrackingItem > File Download Analytics; NPM dependencies update

alexkb0009:
### Part I: Schema for TrackingItem > File Download Analytics

This. Should be reviewed in context Foursight PR https://github.com/4dn-dcic/foursight/pull/278

### Part II: Visualization for this

Not sure if will create another branch for part II but likely.


src/encoded/schemas/tracking_item.json
```diff
@@ -357,8 +373,135 @@
                                     }
                                 }
                             }
+                        },
+
+                        "file_downloads_by_country" : {
+                            "type" : "array",
+                            "title" : "File Downloads",
```
willronchetti, line 380: Extremely minor but make sure spacing is even on all of these, your IDE should show these are off.

PR #1321, opened 2020-05-20
Fix of Column Rendering Problems

utku-ozturk:
1. Very minor fix for `SimpleFilesTable/ItemPageTable`'s `Title` and `Format` columns & `/search/?type=Lab` `P.I.` column
2. `&number_of_experiment_sets.from=1` parameter added to `Browse by Publication` menu link


alexkb0009:
Looks good, I'll give a quick checkout + merge soon, mostly just want to play around with [display: flex](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) stuff a little to see if makes sense to keep [d-flex](https://getbootstrap.com/docs/4.5/utilities/flex/) + maybe see if [flex-direction: column](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox) [(2)](https://getbootstrap.com/docs/4.5/utilities/flex/#direction) would make sense..

PR #1323, opened 2020-05-22
C4-166 CMOD updates

willronchetti:
- On hotseat, skip reindexing altogether
- On staging, do a (strict) full re-index
- Will do the dcicutils refactor as part of this as well, so not spending a lot of time making this look pretty. It will be much better later on.

### DIFF
--- a/src/encoded/commands/create_mapping_on_deploy.py
+++ b/src/encoded/commands/create_mapping_on_deploy.py
@@ -142,10 +142,11 @@ def get_deployment_config(app):
     """
     Gets deployment configuration for the current environment.
 
-    Sets ENV_NAME, WIPE_ES and SKIP as side-effects.
+    Sets ENV_NAME, WIPE_ES, SKIP and STRICT as side-effects.
         ENV_NAME: env we are deploying to
         WIPE_ES: whether or not to completely wipe ES, otherwise only re-index different mappings
         SKIP: whether or not to skip the create_mapping step entirely
+        STRICT: whether or not to "strictly" reindex things (no invalidation)
 
     :param app: handle to Pyramid app
     :return: dict of config options
@@ -155,12 +156,14 @@ def get_deployment_config(app):
     my_env = get_my_env(app)
     deploy_cfg['ENV_NAME'] = my_env
     deploy_cfg['SKIP'] = False  # set to True to skip the create_mapping step
+    deploy_cfg['STRICT'] = False
     if current_prod_env == my_env:
         log.info('This looks like our production environment -- do not wipe ES')
         deploy_cfg['WIPE_ES'] = False
     elif my_env == guess_mirror_env(current_prod_env):
         log.info('This looks like our staging environment -- wipe ES')
         deploy_cfg['WIPE_ES'] = True
+        deploy_cfg['STRICT'] = True
     elif is_stg_or_prd_env(my_env):
         log.info('This looks like an uncorrelated production environment. Something is definitely wrong.')
         exit(0)
@@ -168,6 +171,7 @@ def get_deployment_config(app):
         if is_hotseat_env(my_env):
             log.info('Looks like we are on hotseat -- do nothing to ES')
             deploy_cfg['SKIP'] = True
+            deploy_cfg['STRICT'] = True
         else:
             log.info('Looks like we are on webdev or mastertest -- wipe ES')
             deploy_cfg['WIPE_ES'] = True
@@ -208,11 +212,13 @@ def _run_create_mapping(app, args):
         log.info('Running create mapping on env: %s' % deploy_cfg['ENV_NAME'])
         override_deploy_cfg(deploy_cfg, args.wipe_es, 'WIPE_ES')
         override_deploy_cfg(deploy_cfg, args.skip, 'SKIP')
+        override_deploy_cfg(deploy_cfg, args.strict, 'STRICT')
         if not deploy_cfg['SKIP']:
             run_create_mapping(app,
                                check_first=(not deploy_cfg['WIPE_ES']),
                                purge_queue=args.clear_queue,  # this option does not vary, so no need to override
-                               item_order=ITEM_INDEX_ORDER)
+                               item_order=ITEM_INDEX_ORDER,
+                               strict=deploy_cfg['STRICT'])
     except Exception as e:
         log.error("Exception encountered while gathering deployment information or running create_mapping")
         log.error(str(e))
@@ -229,6 +235,7 @@ def main():
     parser.add_argument('--wipe-es', help="Specify to wipe ES", action='store_true', default=False)
     parser.add_argument('--skip', help='Specify to skip this step altogether', default=False)
     parser.add_argument('--clear-queue', help="Specify to clear the SQS queue", action='store_true', default=False)
+    parser.add_argument('--strict', help='Specify to do a strict reindex', default=False)
 
     args = parser.parse_args()
     app = get_app(args.config_uri, args.app_name)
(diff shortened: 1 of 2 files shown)


j1z0:


```diff

Complexity decreasing per file
==============================
+ src/encoded/commands/create_mapping_on_deploy.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/4dn/fourfront/pullRequest?prid=5571355&bid=18268179)

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -186,13 +207,15 @@ def _run_create_mapping(app, args):
...
-        if args.wipe_es:  # override deploy_cfg WIPE_ES option
-            log.info('Overriding deploy_cfg and wiping ES')
-            deploy_cfg['WIPE_ES'] = True
-        run_create_mapping(app,
-                           check_first=(not deploy_cfg['WIPE_ES']),
-                           purge_queue=args.clear_queue,
-                           item_order=ITEM_INDEX_ORDER)
+        override_deploy_cfg(deploy_cfg, args.wipe_es, 'WIPE_ES')
+        override_deploy_cfg(deploy_cfg, args.skip, 'SKIP')
+        override_deploy_cfg(deploy_cfg, args.strict, 'STRICT')
```
netsettler, line 212: I think this should just be a dictionary
```
apply_dict_overrides(deploy_config, WIPE_ES=args.wipe_es, SKIP=args.skip, STRICT=args.strict)
```
where the following would be in misc_utils, as it would be generally useful in other contexts if it had semantics that were not specific to this app
```
def apply_dict_overrides(dictionary: dict, **overrides) -> dict:
    for k, v in overrides.items():
        if v is not None:
            dictionary[k] = v
    # This function works by side effect, but getting back the changed dict may be sometimes useful.
    return dictionary
```
In other words, I would suggest that if the dictionary entry contains None, it leaves the default and if it contains `True` or `False`, then you would override. This is maximally general for reuse, and yet you can still get the current semantics of your command by just changing the default in the `argparse` specs to be `None` instead of `False`.

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -151,33 +155,50 @@ def get_deployment_config(app):
...
             log.info('Looks like we are on webdev or mastertest -- wipe ES')
             deploy_cfg['WIPE_ES'] = True
     else:
         log.warning('This environment is not recognized: %s' % my_env)
         log.warning('Proceeding without wiping ES')
-        deploy_cfg['WIPE_ES'] = False
     return deploy_cfg
 
 
+def override_deploy_cfg(deploy_cfg, b, key):
```
netsettler, line 182: This would be replaced by `apply_dict_overrides` per my suggestion below.

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -151,33 +155,50 @@ def get_deployment_config(app):
     current_prod_env = whodaman()  # current_prod_end = <current-data-env>
```
netsettler, line 155: You didn't change this part, but thi should use the newer name. I deprecated whodaman. I can fix this when moving  this to utils.

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -151,33 +155,50 @@ def get_deployment_config(app):
...
+    deploy_cfg['STRICT'] = False
+    deploy_cfg['WIPE_ES'] = False
+    if my_env == guess_mirror_env(current_prod_env):
+        log.info('This looks like our staging environment -- wipe ES')
+        deploy_cfg['WIPE_ES'] = True
+        deploy_cfg['STRICT'] = True
     elif is_stg_or_prd_env(my_env):
         log.info('This looks like an uncorrelated production environment. Something is definitely wrong.')
-        exit(0)
+        raise RuntimeError('Tried to run CMOD on production - error\'ing deployment')  # note that this will cause any deployments to production to fail!
```
netsettler, line 167: We should probably make a proper error class to use so that the comment won't be needed.
```
class DeploymentFailure(RuntimeError):
    pass
```

src/encoded/commands/create_mapping_on_deploy.py
```diff
@@ -151,33 +155,50 @@ def get_deployment_config(app):
...
-        deploy_cfg['WIPE_ES'] = False
-    elif my_env == guess_mirror_env(current_prod_env):
-        log.info('This looks like our staging environment -- do not wipe ES')
-        deploy_cfg['WIPE_ES'] = False  # do not wipe ES
+    deploy_cfg['SKIP'] = False  # set to True to skip the create_mapping step
+    deploy_cfg['STRICT'] = False
+    deploy_cfg['WIPE_ES'] = False
+    if my_env == guess_mirror_env(current_prod_env):
+        log.info('This looks like our staging environment -- wipe ES')
+        deploy_cfg['WIPE_ES'] = True
```
netsettler, line 163: This could also use the `apply_dict_overrides` from below, and the set of overrides can be a dictionary with a name like `PRODUCTION_OVERRIDES`, although now that I'm thinking of it, this could also use subclassing, so instead of initializing to `False` and then overriding, one could do:

```
DefaultDeploymentOptions:
    SKIP = False
    STRICT = False
    WIPE_ES = False

ProductionDeploymentOptions(DefaultDeploymentOptions):
    WIPE_ES = True
    STRICT = True

class HotseatDeploymentOptions(DefaultDeploymentOptions):
    SKIP = True
    STRICT = True

class OtherTestDeploymentOptions(DefaultDeploymentOptions):
    WIPE_ES = True
```
and then you do
```
options = GenericDeploymentOptions
if my_env == guess_mirror_env(current_prod_env):  # This function name is deprecated, so we'll want to fix that
    options = ProductionDeploymentOptions
...
```
But after playing it through, i realized it isn't really leveraging the class system enough, so I think just
```
DEFAULT_DEPLOYMENT_OPTIONS = {'SKIP': False, 'STRICT': False, 'WIPE_ES': False}
PRODUCTION_DEPLOYMENT_OPTION_OVERRIDES = {'WIPE_ES': True, 'STRICT': True}
HOTSEAT_DEPLOYMENT_OPTION_OVERRIDES = {'SKIP': True, 'STRICT': True}
OTHER_TEST_DEPLOYMENT_OPTION_OVERRIDES = {'WIPE_ES': True}
```
so that you can do:
```
deploy_cfg.= DEFAULT_DEPLOYMENT_OPTIONS.copy()
...
if my_env == guess_mirror_env(current_prod_env):  # This function name is deprecated, so we'll want to fix that
    deploy_cfg = apply_dict_overrides(deploy_cfg, **PRODUCTION_DEPLOYMENT_OPTION_OVERRIDES)
...
```

src/encoded/tests/test_create_mapping.py
```diff
@@ -62,7 +62,7 @@ def test_get_deployment_config_staging_old():
     """ Tests get_deployment_config in the old staging case """
     cfg = get_deployment_config(None)
     assert cfg['ENV_NAME'] == 'fourfront-webprod'  # sanity
-    assert cfg['WIPE_ES'] is False  # no wipe
+    assert cfg['WIPE_ES'] is True  # wipe
```
netsettler, line 65: Probably should test all three values.

src/encoded/tests/test_create_mapping.py
```diff
@@ -72,27 +72,25 @@ def test_get_deployment_config_staging_new():
     """ Tests get_deployment_config in the new staging case """
     cfg = get_deployment_config(None)
     assert cfg['ENV_NAME'] == 'fourfront-blue'  # sanity
-    assert cfg['WIPE_ES'] is False  # no wipe
+    assert cfg['WIPE_ES'] is True  # wipe
```
netsettler, line 75: Probably should test all three values.

PR #1325, opened 2020-05-26
C4-167 PURGE fix

willronchetti:
- Hooks in latest snovault, which fixes item PURGE on production


pyproject.toml
```diff
@@ -44,7 +44,7 @@ chardet = "3.0.4"
 "collective.recipe.modwsgi" = "2.1"
 "collective.recipe.template" = "1.13"
 colorama = "0.3.3"
-dcicsnovault = ">=3.0.0,<4"
+dcicsnovault = ">=3.0.2,<4"
```
netsettler, line 47: Probably only re-lock was needed but given you were already in there, this can't hurt.

willronchetti: My point here was to make `poetry install` fail with snovault prior to this fix (don't want any prior version 3 installs).

PR #1346, opened 2020-07-07
Rewrite get_item_if_you_can as get_item_or_none to match CGAP

netsettler:
In CGAP, we renamed `get_item_if_you_can` to `get_item_or_none`. This does the same for Fourfront so that diffs of the two repos aren't gratuitously different in a bunch of places.

That's all this PR does. Nothing else riding along, so should be easy to review. :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,10 +78,10 @@ clean-python:
 	pip freeze | xargs pip uninstall -y
 
 test:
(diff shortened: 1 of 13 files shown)


netsettler:
There were problems in testing but largely they came down to wrong time of day for testing. There is definitely some bad transient afoot. In any case, I think this edit is fine.

PR #1349, opened 2020-07-08
Replicate Set Overview for Experiments That Lack Files

utku-ozturk:
**Problem (from Trello):** 
replicate set overview for experiments that lack files (neither raw nor processed). eg. imaging experiments (e.g. https://data.4dnucleome.org/browse/?type=ExperimentSetReplicate&experimentset_type=replicate&experiments_in_set.files.file_format.display_title=No+value&experiments_in_set.processed_files.file_format.display_title=No+value)

**Solution:**
A new table compatible w/ processed file table is added:

Screen Shot 2020-07-08 at 21 33 06


utku-ozturk:
@alexkb0009 refactored the code and memoized both header row and exps not associated w/ files row.

PR #1350, opened 2020-07-09
Upgrade to coverage 5.1 so that coveralls can work (C4-218)

netsettler:
As with CGAP, coveralls has long been broken on Travis because they dropped support for coverage 4.x and require an upgrade to version 5. This fixes that. [It's broken right now on master](https://travis-ci.org/github/4dn-dcic/fourfront/jobs/706267907) and [is fixed on this branch](https://travis-ci.org/github/4dn-dcic/fourfront/jobs/706488366).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "1.7.15"
+version = "1.7.16"
 description = "4DN-DCIC Fourfront"
 authors = ["4DN-DCIC Team <support@4dnucleome.org>"]
 license = "MIT"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -117,7 +117,7 @@ xlwt = "1.2.0"
 "zope.sqlalchemy" = "^1.2"
 
 [tool.poetry.dev-dependencies]
-coverage = "4.0.3"
+coverage = "5.1"
```
willronchetti, line 120: On CGAP the mark was >=5.1, is there a reason for the discrepancy here?

netsettler: Nope. Careless on my part. I'll fix that.

netsettler: In fact that will get coverage 5.2, but that's good enough and is what CGAP is proposed to do. It's up testing now. I'll merge it when it goes green.

PR #1357, opened 2020-07-28
add "grouped with" file relation

andreacosolo:
A generic, non-hierarchical relation between files. Useful for grouping.


src/encoded/types/file.py
```diff
@@ -571,7 +571,8 @@ def _update(self, properties, sheets=None):
             "parent of": "derived from",
             "supercedes": "is superceded by",
             "is superceded by": "supercedes",
-            "paired with": "paired with"
+            "paired with": "paired with",
+            "grouped with": "grouped with"
```
aschroed, line 575: Did you add inserts with these relationships?  Also did you test to see if the relationships were made reciprocally when only one was provided in an excel spreadsheet loaded by submit4DN - locally?

PR #1364, opened 2020-08-04
C4-248 Integrate Sentry

willronchetti:
- Integrates the sentry SDK to our application
- Patches C4-223 while we're at it


pyproject.toml
```diff
@@ -115,6 +115,7 @@ xlwt = "1.2.0"
 "zope.deprecation" = "4.4.0"
 "zope.interface" = "4.6.0"
 "zope.sqlalchemy" = "^1.2"
+sentry-sdk = "^0.16.2"
```
netsettler, line 118: Uparrow does little in version 0, only changes patch version. Maybe want `">=0.16.2"` or `">=0.16.2,<1"`. The latter is what you probably thought you meant. But really if we just want their highest version, probably we shouldn't even do `"<1"`.

ebberg:
Belated, but looks good to me.

PR #1369, opened 2020-08-12
C4-268 Configure ES Cluster from EB

willronchetti:
- Derive ES cluster location from EB configuration, not from the `ini` files.


deploy/ini_files/any.ini
```diff
@@ -7,7 +7,7 @@ blob_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-blobs
 system_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-system
 # blob_store_profile_name = encoded-4dn-files
 accession_factory = encoded.server_defaults.enc_accession
-elasticsearch.server = ${ES_SERVER}
+elasticsearch.server = ${ES_URL}
```
netsettler, line 10: The `ES_SERVER` environment variable here that you're perhaps wondering where it comes from is created automatically by deployment utils either from specifying `ENCODED_ES_SERVER` in the instance or from passing a `es_server` in the deploy command.  I think you should arrange to set one of those and pick it up as `ES_SERVER` here so you don't have to change deployment_utils.

willronchetti: I've updated all to use `ES_SERVER` and have set `ENCODED_ES_SERVER` on the relevant environments.

PR #1372, opened 2020-08-24
Embed qc fields

Lmercadom:


src/encoded/types/experiment_set.py
```diff
@@ -218,6 +218,8 @@ class ExperimentSet(Item):
         "processed_files.quality_metric.overall_quality_status",
         "processed_files.quality_metric.quality_metric_summary.*",
         "processed_files.notes_to_tsv",
+        "processed_files.quality_metric.Total reads",
```
netsettler, line 221: Really there's no space in the property name? Seems like it would be so much better as `total_reads`, just in case someone needs to use the property name as a variable name.

netsettler: Though I guess the property name was not created by this PR. Sigh.

Lmercadom: Yeah, at first they were free fields that were calculated, but then after the refactoring, they became actual properties so it could not be changed :/

PR #1373, opened 2020-08-26
Snovault 3.2.0

willronchetti:
- Hook in snovault 3.2.0, allowing calculated properties on sub-embedded object fields and arrays of sub-embedded objects.


pyproject.toml
```diff
@@ -94,7 +94,7 @@ rutter = ">=0.2,<1"
 s3transfer = "^0.2.0"
 simplejson = "^3.17.0"
 SPARQLWrapper = "1.7.6"
-SQLAlchemy = "1.2.16"
+SQLAlchemy = "1.3.16"
```
netsettler, line 97: I'm quite surprised this is not already set. I thought I'd put in a PR about that. Please change this line to add the comment
```
SQLAlchemy = "1.3.16"  # Pinned because >=1.3.17 is a problem
```
so that it matches what's in cgap-portal. Likewise please change this line:
```
"zope.sqlalchemy" = "1.3"  # Believed to work well with sqlalchemy 1.3.16
```

PR #1387, opened 2020-10-05
C4-325 Fix Deploy

willronchetti:
- Ported from CGAP
- On 10/1 a change was introduced to EB that triggered our setup_eb.py script to generate an extra directory and .pth file breaking the deployment
- Fix this by removing these files with 0698_remove_extra_files
- Also pins pip and setuptools versions


netsettler (approved):
You need to bump the version number. And I'd like the same comments I requested in [CGAP PR #226](https://github.com/dbmi-bgm/cgap-portal/pull/226). But otherwise this is fine.

PR #1392, opened 2020-10-09
Update snovault

willronchetti:
- Updates snovault version with some performance fixes in indexing.


netsettler (commented):
It doesn't let me formally approve this since it's already merged, but I do approve.

PR #1399, opened 2020-10-19
Pin Poetry Version

willronchetti:
- Provides "true" fix to #1387 by pinning a working poetry version in the deployment

### DIFF
--- a/.ebextensions/20_packages.config
+++ b/.ebextensions/20_packages.config
@@ -21,7 +21,7 @@ container_commands:
   0191_check_pip_version_anew:
     command: "source /opt/python/run/venv/bin/activate && pip --version"
   0200_install_poetry:
-    command: "source /opt/python/run/venv/bin/activate && pip install poetry >> /var/log/deploy.log"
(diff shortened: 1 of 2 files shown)


.ebextensions/20_packages.config
```diff
@@ -50,8 +50,6 @@ container_commands:
     command: "source /opt/python/run/venv/bin/activate && python deploy/generate_production_ini.py"
   0696_pip_install_encoded:
     command: "source /opt/python/run/venv/bin/activate && python setup_eb.py develop --verbose"
-  0698_remove_extra_files:  # XXX: This is required as of 10/1/2020 to remove files that mess up the install -Will
-    command: "rm /opt/python/run/venv/lib/python3.6/site-packages/encoded.pth && rm -rf /opt/python/run/venv/lib/python3.6/site-packages/encoded-[0-9]*.dist-info"
```
netsettler, line 54: I could see us having to undo the pinning before we have a better solution, so I somewhat think we should just disable and not remove this. It's a pain to do archaeology to find such things..

netsettler (approved):
My comment is not blocking.

PR #1402, opened 2020-10-22
C4-356 Disable unauthenticated downloads

willronchetti:
- Raise HTTPForbidden if a file download is attempted while not logged in.
- Ported from https://github.com/dbmi-bgm/cgap-portal/pull/245


netsettler (approved):
See my style comments on [the similar PR (245) in cgap-portal](https://github.com/dbmi-bgm/cgap-portal/pull/245).
No blockers though.

PR #1403, opened 2020-10-26
Fileformat metadata

sbreiff:
The purpose of these changes are to add more properties to FileFormat items so that we can build a File Formats Resources page that uses an Embedded Search Table with helpful information to users.

**Changes:**
* `assay_types`, `file_content`, `visualization`, `url` properties added to `file_format.json` schema
* facets and columns updated on `file_format.json` schema
* when `url` property is used as a column in a search table, it now appears as an external link instead of plain text
* file_format master inserts updated with some of the new properties. Some descriptions added/edited but more additions/edits can just be done on DB items.


alexkb0009 (approved):

src/encoded/schemas/file_format.json
```diff
@@ -127,22 +127,106 @@
...
+            "items": {
+                "title": "Assay or Category",
+                "type": "string",
+                "suggested_enum": [
+                    "Sequencing",
+                    "Microscopy",
+                    "Hi-C",
+                    "SPRITE",
+                    "RNA-seq",
+                    "Generic"
```
aschroed, line 144: Is this a catchall?  What format(s) might include generic and how would that fit into the result table?

I realize that this is a suggested enum and I know you've been thinking about this - I just don't have a good sense as to if you think these values currently will be sufficient or is the plan to add more as you work on the look of the embedded search table?

src/encoded/schemas/file_format.json
```diff
@@ -127,22 +127,106 @@
...
+                "alignments",
+                "contact list",
+                "contact matrix",
+                "image",
+                "intermediate file",
+                "multi-contact read clusters",
+                "other",
+                "reads",
+                "sequences",
+                "SPT results",
```
aschroed, line 165: Might this be better as something more general like 'particle position or trajectory results'?

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "2.1.8"
+version = "2.2.7"
```
aschroed, line 4: Bumping the minor version but decrementing the patch version by one?  I'm not sure why not just 2.2.0 or if no real incompatibilities 2.1.9?

aschroed (approved):
A couple of minor comments which aren't blocking and may not be relevant.

src/encoded/schemas/file_format.json
sbreiff: As far as “Generic” in the file_formats assay type, this is meant for formats like “txt” meaning that it isn’t for a particular assay type but is a generic file format. The other option would be to remove that from the enum and leave that field blank for those file formats.

As to whether the values in the enum are sufficient, i won’t be surprised if changes are needed later on. Maybe replacing “SPRITE” with “multi-contact assays” or something could make sense. For our current purposes though I felt the values were sufficient - 1D sequencing experiments use formats like bed and bw and I think “Sequencing” is an appropriate category for that, without specifying what kind of sequencing.

src/encoded/schemas/file_format.json
sbreiff: Since this is meant to appear in a table column 'particle position or trajectory results' would be too long to be useful I think but I can think about whether I can come up with an alternative.

pyproject.toml
sbreiff: Ah you're right, this was a mistake. I'll fix that.

sbreiff: But incompatibilities are only relevant for a major version increase, as far as I'm aware. New features merit a minor version increase, and patch version increases are mainly for bug fixes. I wanted to increase the minor version because the url handling in the search tables was a new feature, IMO.

PR #1407, opened 2020-11-05
File Download Column for Protocols Search Table

utku-ozturk:
**Trello Ticket:** https://trello.com/c/9qULcpwx

**Solution:**
The new "attachment" column with "File" as title added to `Protocols` schema and custom rendering is added to `columnExtensionMap` to render the pdf icon download link.

Screen Shot 2020-11-06 at 00 56 39


alexkb0009 (approved):
Awesome!

**Edit**: There's failing travis tests but these seem very unrelated. ~~Will trigger a re-run...~~ maybe we need to merge in master for fixes?

PR #1408, opened 2020-11-09
C4-343 Additional Facets

willronchetti:
- Ported from CGAP: https://github.com/dbmi-bgm/cgap-portal/pull/253

### DIFF
--- a/src/encoded/search.py
+++ b/src/encoded/search.py
@@ -982,27 +982,26 @@ def initialize_facets(request, doc_types, prepared_terms, schemas, additional_fa
     initialize_additional_facets(request, doc_types, additional_facets,
                                  append_facets, current_type_schema)
 
-
     # Add facets from schema if one Item type is defined.
     # Also, conditionally add extra appendable facets if relevant for type from schema.
     if len(doc_types) == 1 and doc_types[0] != 'Item':
         if 'facets' in current_type_schema:
             schema_facets = OrderedDict(current_type_schema['facets'])
             for schema_facet in schema_facets.items():
-                if schema_facet[1].get('disabled', False):
+                if schema_facet[1].get('disabled', False) or schema_facet[1].get('default_hidden', False):
                     disabled_facets.append(schema_facet[0])
-                    continue # Skip disabled facets.
+                    continue  # Skip disabled facets.
                 facets.append(schema_facet)
 
-    ## Add facets for any non-schema ?field=value filters requested in the search (unless already set)
-    used_facets = [ facet[0] for facet in facets + append_facets ]
+    # Add facets for any non-schema ?field=value filters requested in the search (unless already set)
+    used_facets = [facet[0] for facet in facets + append_facets]
     used_facet_titles = used_facet_titles = [
         facet[1]['title'] for facet in facets + append_facets
         if 'title' in facet[1]
     ]
     for field in prepared_terms:
         if field.startswith('embedded'):
-            split_field = field.strip().split('.') # Will become, e.g. ['embedded', 'experiments_in_set', 'files', 'file_size', 'from']
+            split_field = field.strip().split('.')  # Will become, e.g. ['embedded', 'experiments_in_set', 'files', 'file_size', 'from']
             use_field = '.'.join(split_field[1:])
 
             # 'terms' is the default per-term bucket aggregation for all non-schema facets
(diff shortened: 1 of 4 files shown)


src/encoded/search.py
```diff
@@ -494,9 +495,9 @@ def prepare_search_term(request):
     Ignore certain keywords, such as type, format, and field
     """
     prepared_terms = {}
-    prepared_vals = []
-    for field, val in request.normalized_params.iteritems():
-        if field.startswith('validation_errors') or field.startswith('aggregated_items'):
+    for field, val in request.normalized_params.items():
+        if field.startswith('validation_errors') or field.startswith('aggregated_items') or \
```
netsettler, line 499: maybe factor this test out as a predicate (`is_additional_facet_name` or some such?)

src/encoded/search.py
```diff
@@ -898,7 +899,54 @@ def set_filters(request, search, result, principals, doc_types):
...
+        :param append_facets: list to add additional_facets to
+        :param current_type_schema: schema of the item we are faceting on
+    """
+    for extra_facet in additional_facets:
+        aggregation_type = 'terms'  # default
+
+        # check if defined in facets
+        if 'facets' in current_type_schema:
+            if extra_facet in current_type_schema['facets']:
+                if not current_type_schema['facets'][extra_facet].get('disabled', False):
```
netsettler, line 917: There's no advantage of specifying `False` here in place of the default `None` since you are in a boolean test context anyway. Missing from `.get(...)` is pretty well-accepted as false-y so the extra emphasis doesn't add much. Fine either way though.

src/encoded/search.py
```diff
@@ -930,10 +978,14 @@ def initialize_facets(request, doc_types, prepared_terms, schemas):
     # hold disabled facets from schema; we also want to remove these from the prepared_terms facets
     disabled_facets = []
 
+    current_type_schema = request.registry[TYPES][doc_types[0]].schema
```
netsettler, line 981: Is the first the default? I don't like seeing array references inside array references. Maybe it's a worry one will fail and you won't know which one. But it's also multiple abstractions in play at once. I'd be happier with (untested):
```
default_doc_type = doc_types[0]
current_type_schema = request.registry[TYPES][default_doc_type].schema
```
or something like that.

src/encoded/search.py
```diff
@@ -930,10 +978,14 @@ def initialize_facets(request, doc_types, prepared_terms, schemas):
...
+    initialize_additional_facets(request, doc_types, additional_facets,
+                                 append_facets, current_type_schema)
+
+
     # Add facets from schema if one Item type is defined.
     # Also, conditionally add extra appendable facets if relevant for type from schema.
     if len(doc_types) == 1 and doc_types[0] != 'Item':
```
netsettler, line 988: Seems like you've already presumed the doc_types is length 1 above, no? Anyway, if that logic were done as I suggested above could you still use that variable here?

src/encoded/tests/test_search.py
```diff
@@ -642,3 +642,227 @@ def test_barplot_aggregation_endpoint(workbook, testapp):
...
+            else:
+                assert facet['aggregation_type'] == 'terms'
+
+    @pytest.mark.parametrize('_facets', [
+        ['last_name', 'unfaceted_integer'],  # second slot holds number field
+        ['unfaceted_string', 'sid']
+    ])
+    def test_search_mixing_additional_and_default_hidden(self, testapp, hidden_facet_test_data, _facets):
+        """ Tests that we can mix additional_facets with those both on and off schema """
+        facets = testapp.get('/search/?type=TestingHiddenFacets'
```
netsettler, line 789:
```
[sample_string_field, sample_number_field] = _facets
```
would then allow you to reference them as other than just `_facets[0]` and `_facets[1]`.

src/encoded/tests/test_search.py
```diff
@@ -642,3 +642,227 @@ def test_barplot_aggregation_endpoint(workbook, testapp):
...
+                             '&additional_facet=%s'
+                             '&additional_facet=%s' % (_facets[0], _facets[1])).json['facets']
+        expected = self.DEFAULT_FACETS + _facets
+        actual = [facet['field'] for facet in facets]
+        assert sorted(expected) == sorted(actual)
+        for facet in facets:
+            if facet['field'] == _facets[1]:  # second slot holds number field
```
netsettler, line 796: If `_facets[1]` were `sample_number_field` or some such, then it wouldn't need this comment.

netsettler (approved):
I did a slightly lighter read on this than I did on the cgap one because it seemed more familiar, but it looks basically fine in spite of my non-blocking comments.

PR #1411, opened 2020-11-12
Metadata.tsv Edits

utku-ozturk:
From Trello ([this](https://trello.com/c/iO6vT9rN) and [this](https://trello.com/c/dZ0hfGOB)):

1. **Bugfix on file relationships (ProcessedFiles)+ improvement of relationship description (order of relationships)**: 

While playing around with sets that have many relationships among files (example), I found that in metadata.tsv they are sometimes misrepresented, and sometimes even completely ignored. In particular:

- When file A is grouped with B, grouped with C, parent of D, the Related File Relationship column is “grouped with, parent of” and Related File is “B, C, D”. This is confusing to me. The ideal result would be: Related File Relationship = “grouped with, grouped with, parent of”, and Related File is “B, C, D” (in the correct order).
- The FileProcessed items in this example have a “derived from” relationship, which is not shown in the metadata.tsv. I don’t know if this is a bug related to this specific relationship type (or a feature?), or if it has something to do with FileProcessed. Anyway, the expected result would be to see also these relationships in the tsv file.

2. **Experiment Reference File(s)**:

Add reference_files to metadata.tsv (only in certain cases): Experiments sometimes have files listed under `reference_files` property. These are currently not included in the metadata.tsv. We would like to include them, but only if the selection of files to download includes at least one raw file from that experiment. Conversely, if one downloads only processed files, the reference_files are not interesting and should not be downloaded.

3. **OPF Downloads**

In the ExperimentSet page, both Raw Files and Processed Files tabs have each a download button. Add the same also to the Supplementary Files tab.


alexkb0009 (approved):
Looks+works great!
I really like how the `SelectedFilesController` (+ UX/button) wraps/applies-to all the SPF collections (my first thought would have been to wrap each 1 of them, but that would have been worse I think).

PR #1413, opened 2020-11-16
C4-370 Range Facets

willronchetti:
- Ports https://github.com/dbmi-bgm/cgap-portal/pull/259

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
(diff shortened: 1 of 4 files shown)


netsettler (approved):
i tried to compare this to the CGAP code, but CGAP has substantially diverged due to nesting. We really need to figure out if that can't be unified somehow. That's a lot of dissonance to maintain. For the purposes of this patch, it only means that it's hard to use the prior PR as a guide.

Nothing seemed obviously wrong here, though, and the tests at least aligned, so I assume we didn't broke anything and we've added new functionality.

I don't see any blockers.

PR #1414, opened 2020-11-22
added external_submission

andreacosolo:
Created an `external_submission` mixin property and added that to ExpSet, Exp and Biosample.
When an item is submitted by 4DN DCIC to an external resource, the database name and date of export will be recorded here.


netsettler (commented):
This seems not to have had a reviewer, but I looked at it after-the-fact and it seemed OK.

It should have been a minor version bump, but that's ... well, ... minor.

andreacosolo:
It had voice approval from Andy.

PR #1415, opened 2020-11-23
update biosource_summary

andreacosolo:
Only add "differentiated to" if in_vitro_differentiated == Yes


src/encoded/types/biosample.py
```diff
@@ -194,7 +194,8 @@ def biosource_summary(self, request, biosource, cell_culture_details=None):
             ret_str = ret_str[:-5]
             if cell_culture_details:  # will assume same differentiation if multiple bccs
                 cc_props = get_item_or_none(request, cell_culture_details[0], 'biosample_cell_cultures', frame='embedded')
-                if cc_props and 'tissue' in cc_props:
+                if (cc_props and 'tissue' in cc_props and
+                        cc_props.get('in_vitro_differentiated', '') == 'Yes'):
```
aschroed, line 198: Maybe better to:
cc_props.get('in_vitro_differentiated', 'No')

aschroed (approved):
Looks good.  Non-blocking suggestion - use the default value in the get rather than empty string?

PR #1416, opened 2020-12-10
Added molecule to ExperimentSeq to specify what is sequenced (especially useful for RNA-seq)

andreacosolo:


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "2.3.7"
+version = "2.4.0"
```
sbreiff, line 4: I feel like this could be a patch version bump instead, but it probably doesn't matter that much.

sbreiff (approved):

PR #1417, opened 2020-12-11
ExperimentSetsTableTabView: Allow to look up and render static_sections for location "tab:expsets-table"

utku-ozturk:
Trello ticket: https://trello.com/c/HOf6S4bd

Solution (screenshot):

Screen Shot 2020-12-11 at 12 36 22


Bianca-Morris (approved):
Looks good to me.

src/encoded/static/components/item-pages/components/TabbedView.js
```diff
@@ -49,7 +49,20 @@ export function getTitleForCustomTab(tabName){
...
+    if (context && context.static_content && Array.isArray(context.static_content) && typeof tabLocation === 'string') {
+        const staticContent = _.pluck(_.filter(context.static_content || [], function (s) {
+            return s.content && !s.content.error && s.location === tabLocation;
+        }), 'content');
+        return staticContent;
+    }
 
+    return null;
+});
+export { getTabStaticContent };
```
alexkb0009, line 65: Would remove this `export { getTabStaticContent };` and instead try to have export precede the const definition, e.g. ~ `export const getTabStaticContent = memoize(function (context, tabLocation) { ... ` for consistency w. other exports.

Would consider re-using this function in places where this functionality is repeated, namely [PublicationView.js](https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/static/components/item-pages/PublicationView.js#L229-L231) and [ExperimentTypeView](https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/static/components/item-pages/ExperimentTypeView.js#L76-L78). This function could also be made to return an empty array if no staticContent found -- this is mostly subjective (& allows to not prepend `staticContent &&` [here](https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/static/components/item-pages/PublicationView.js#L246)) but might be good to only have 1 return type, idk.

(Not too important now:)
For future, if we want to reuse this function more (&/or have more static content on more tabs of ItemViews), would be good to just `export function getTabStaticContent(context, tabLocation) { ..` (without _globally_ memoizing it). And then perhaps have any component that uses the func, to memoize its own version/copy (or wrap it with a `useMemo(function(){ return getTabStaticContent(context, '..') }, [ context ])` if is a functional component). This would be useful because multiple `getTabStaticContent` calls could potentially be executed within a single view, as tabs' contents are cached in the DOM (the element(s)/components containing the tab is not unmounted from the DOM) so code that renders other inactive Tabs in a given ItemView might run alongside code that renders the active Tab. This would also allow this function to be [re-used in TabbedView.js itself](https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/static/components/item-pages/components/TabbedView.js#L163-L165).

alexkb0009 (changes requested):
Checkout comment that left re: getTabStaticContent; would try to add in what mentioned in 1st & 2nd paragraph and keep 3rd paragraph in mind for future if need to add static sections to more tab.

PR #1420, opened 2020-12-16
Workaround for deploy1 bug (C4-392) and update dependencies to mirror cgap-portal

netsettler:
This does these things:

* Fix [make deploy1 is broken on cgap-portal and fourfront (C4-392)](https://hms-dbmi.atlassian.net/browse/C4-392) by pinning the version of poetry in `make configure`.
* Update Makefile to be more like in `cgap-portal`:
   * More variations of `make clean`.
   * More variations of `make build`.
   * Add `make npm-setup-if-needed`.
* Update a number of dependency version constraints in `pyproject.toml`.


pyproject.toml
```diff
@@ -35,18 +35,19 @@ classifiers = [
...
-"collective.recipe.template" = "1.13"
+# I think .recipe is a buildout thing so I'm going to try commenting this out as a start. Delete later. -kmp 15-Dec-2020
+# "collective.recipe.cmd" = "0.11"
+# "collective.recipe.modwsgi" = "2.1"
+# "collective.recipe.template" = "1.13"
 colorama = "0.3.3"
-dcicsnovault = "^4.0.12"
-dcicutils = "^1.3.0"
+dcicsnovault = "4.1.0"
+dcicutils = "^1.4.0"
```
willronchetti, line 50: Is there a reason we aren't aligning to a higher version? Is it because of the library update for `elasticsearch`? It should be ok to bump that as well.

willronchetti (approved):

pyproject.toml
netsettler: At this point I'm just trying to match what we did in CGAP, which had this pinned. It seemed unlikely that Fourfront needs a newer version than CGAP.

PR #1421, opened 2020-12-17
Add date modified to optional columns

utku-ozturk:
**Trello:** https://trello.com/c/7Sd6XvQ5
**Burak:** I always find myself wanting to sort by date modified. It should be available for all searches.


Bianca-Morris (commented):
Sort seems to work well on the items I've tried. The text-alignment of that new column [seems a bit off](https://gyazo.com/f7be67ef1e6e8e424c584bd362186ca3) compared to the others, so you might need to add a line in [_search.scss](https://github.com/4dn-dcic/fourfront/blob/b77753123af0ca30b0952713694c46ef3e6a42aa/src/encoded/static/scss/encoded/modules/_search.scss#L85) or maybe [fourfront's colExtensionMap](https://github.com/4dn-dcic/fourfront/blob/add_date_modified/src/encoded/static/components/browse/columnExtensionMap.js) in order to get that center-aligned. Kind of weird that it's not doing that by default; I thought it had in the past... might be something Alex or I messed up while working on CGAP. Might also be worth checking embedded search tables for the same issue if those are present anywhere on fourfront (can't remember off the top of my head if they are; you'd know better than I). Otherwise looks good!

alexkb0009 (approved):
Awesome! Agree w. what Bianca said + then good to merge.

Best to do the colExtensionMap approach and clone [this column definition (including render function)](https://github.com/4dn-dcic/fourfront/blob/add_date_modified/src/encoded/static/components/browse/columnExtensionMap.js#L167) (or [this one](https://github.com/4dn-dcic/shared-portal-components/blob/master/src/components/browse/components/table-commons/basicColumnExtensionMap.js#L67) - same thing) but with "last_modifed.date_modifed" as the key and value used from 'result'.

If we want this for both CGAP and 4DN, then could probably merge this PR now though and define the column render func in [this SPC file](https://github.com/4dn-dcic/shared-portal-components/blob/master/src/components/browse/components/table-commons/basicColumnExtensionMap.js) instead of the 1 in this repo.

alexkb0009 (changes requested):
See prev. comment

utku-ozturk:
> Awesome! Agree w. what Bianca said + then good to merge.
> 
> Best to do the colExtensionMap approach and clone [this column definition (including render function)](https://github.com/4dn-dcic/fourfront/blob/add_date_modified/src/encoded/static/components/browse/columnExtensionMap.js#L167) (or [this one](https://github.com/4dn-dcic/shared-portal-components/blob/master/src/components/browse/components/table-commons/basicColumnExtensionMap.js#L67) - same thing) but with "last_modifed.date_modifed" as the key and value used from 'result'.
> 
> If we want this for both CGAP and 4DN, then could probably merge this PR now though and define the column render func in [this SPC file](https://github.com/4dn-dcic/shared-portal-components/blob/master/src/components/browse/components/table-commons/basicColumnExtensionMap.js) instead of the 1 in this repo.

Thanks for the feedbacks. Currently, I have only extended the columnExtensionMap. The new look and feel is: (I'll merge it shortly)
Screen Shot 2021-01-07 at 17 29 36

src/encoded/static/components/browse/columnExtensionMap.js
```diff
@@ -230,4 +230,31 @@ export const columnExtensionMap = _.extend({}, basicColumnExtensionMap, {
     'experiments_in_set.biosample.biosource_summary' : {
         'widthMap' : { 'lg' : 260, 'md' : 230, 'sm' : 180 }
     },
+    'date_created' : {
```
alexkb0009, line 233: We can remove `date_created` from here since this extends basicColumnExtensionMap from SPC which has it already.

utku-ozturk: @alexkb0009  date_created of basicColumnExtensionMap is right-aligned but here we override it as center-aligned. if we remove it, Created column and Modifided columns' styles are not matching.

alexkb0009: Good point, sorry didn't pay much attention re: the 'centering' part in prev comments --

I think the most important thing above all here is to have all date columns render in the same exact style. Whether that means is text-center or text-right, I guess is somewhat subjective and not that important.

But if we're gonna change to have these 2 date columns be `text-center` instead of `text-right`, to stay consistent, we should change this for _all_ columns, including `google_analytics.for_date` and `date_published ` within this file as well as for the date fields in SPC, and the ones in CGAP.

(I'd argue that keeping all dates be `text-right` would be preferable, both because less code changes needed from current state, and (subjective/arguably-) aligning the year might help to see differences in date more easily when many rows are shown)

PR #1424, opened 2021-01-20
Babel Config Update

alexkb0009:
Fixes NPM/Jest test runs, [code-splitting is still working as expected](https://gyazo.com/a7be9780a48724dc8027a0b28fa5b78e) (output of `npm run dev-analyzed`, which ran in 'development' environment).

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,7 +1,7 @@
 
 module.exports = function(api){
     const modulesEnabled = api.env("test");
-    api.cache(true);
+    api.cache.using(function(){ return process.env.NODE_ENV; });
     return {
         "presets" : [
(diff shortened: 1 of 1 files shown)


andreacosolo (approved):

babel.config.js
```diff
@@ -1,11 +1,12 @@
 
 module.exports = function(api){
-    api.cache(true);
```
alexkb0009, line 3: Was equivalent to `api.cache.forever()`, which not meant to be used alongside [api.env](https://babeljs.io/docs/en/config-files#apienv) in its doc note.

utku-ozturk (approved):
tested locally, works great!

PR #1430, opened 2021-02-02
Upgrading Cypress Version to 6.2.1

utku-ozturk:
Pls also refer to: https://github.com/4dn-dcic/post-deploy-cypress-tests/pull/4

- FF's existing Cypress version was `4.10.0`. Now it is updated as `6.2.1`, which is the most recent stable version.
- There is a breaking change compared to the old version: `blacklistHosts` is renamed as `blockHosts`. (https://docs.cypress.io/guides/references/migration-guide.html#blacklistHosts-configuration-renamed).
- The new version was run successfully against local, staging, and data.


alexkb0009 (approved):
Didnt test on local but looks like it should work perfectly.

PR #1431, opened 2021-02-03
Chg tr unit to sugg enum

aschroed:
Changed treatment_agent.units enum to suggested_enum
Added pfu to units.

### DIFF
--- a/src/encoded/schemas/treatment_agent.json
+++ b/src/encoded/schemas/treatment_agent.json
@@ -81,7 +81,7 @@
             "title": "Concentration units",
             "type": "string",
             "lookup": 40,
-            "suggest_enum": [
+            "suggested_enum": [
                 "pM",
                 "nM",
                 "μM",
--- a/src/encoded/schemas/treatment_agent.json
+++ b/src/encoded/schemas/treatment_agent.json
@@ -81,7 +81,7 @@
             "title": "Concentration units",
             "type": "string",
             "lookup": 40,
-            "enum": [
+            "suggest_enum": [
                 "pM",
                 "nM",
                 "μM",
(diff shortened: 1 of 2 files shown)


src/encoded/schemas/treatment_agent.json
```diff
@@ -93,7 +93,8 @@
                 "percent",
                 "units",
                 "U/mL",
-                "MOI"
+                "MOI",
+                "pfu"
```
andreacosolo, line 97: I'd suggest to write PFU in capital letters, to be consistent with MOI (or both small?). Unless this is the way they are most commonly used (I'm not too familiar with these units).

andreacosolo (approved):
Just a non-blocking comment on notation.

src/encoded/schemas/treatment_agent.json
```diff
@@ -81,7 +81,7 @@
             "title": "Concentration units",
             "type": "string",
             "lookup": 40,
-            "enum": [
+            "suggest_enum": [
```
sbreiff, line 84: doesn't this need to be "suggested_enum"?

src/encoded/schemas/treatment_agent.json
sbreiff: both capital, IMO.

src/encoded/schemas/treatment_agent.json
aschroed: Of course it does - this is exactly why we do reviews even for small changes like this.  Doh! Thanks. Fixed now I hope.

sbreiff (changes requested):
change suggest_ to suggested_. [EDIT: you've just done this so I approve.] I also agree with Andrea about the capitals.

One other question though - is it total PFU? Or PFU/ml? These are concentration units and "PFU" isn't a concentration.

sbreiff (approved):

PR #1433, opened 2021-02-10
Publication Experiment Sets Tab

utku-ozturk:
**Trello:** Publication ExperimentSets tab all experiment sets are listed in one table - and in some cases, it's useful and somewhat important to split them into two classes based on how they are linked to the publication `expsets_produced_in_pub` or `expsets_used_in_pub` ([card](https://trello.com/c/hbIVAa6t))

**Solution:**

First of all, we added a new calculated property `pubs_using` into the `ExperimentSet` item to search experiment sets used in publications.

``` python
    @calculated_property(schema={
        "title": "Publications Using",
        "description": "Publications using this Experiment Set",
        "type": "array",
        "items": {
            "title": "Publication",
            "type": "string",
            "linkTo": "Publication"
        }
    })
    def pubs_using(self, request):
        return self.rev_link_atids(request, 'publications_using')
```

Note: there already exists a calculated property `produced_in_pub` for the publication in which this Experiment Set was produced: https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/types/experiment_set.py#L328-L339

After splitting the table into two separate tables, then the new look and feel turned into the screenshot below:

Screen Shot 2021-02-10 at 22 56 56


alexkb0009 (approved):
Looks great! (DIdn't test on local)

aschroed:
Question - would it be difficult to add a near the top with the total number of sets total?  For the example linked on hotseat it would say something like '56 Experiment Sets associated with this Publication'.  There was a concern that for small screens one might never realize or see that there is the second table

utku-ozturk:
@aschroed just updated hotseat: 

- display a descriptive title (`XX Experiment Sets associated with this Publication`) if there are multiple tables
- moved subtitles (`Produced In Publication` & `Used By Publication`) next to results count so each table section is more compact now.

Firefox_Screenshot_2021-02-12T13-24-48 644Z

aschroed:
Looks great to me.  Thanks!

PR #1434, opened 2021-02-13
Update Fourfront to get GA configuration to work again (C4-600)

netsettler:
This tries to fix bug C4-600 by doing an `apt update` before installing Postgres.
That seems to get past the problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Install/Link Postgres
         run: |
-          sudo apt update
+          sudo apt update -yq
           sudo apt install -yq --no-install-suggests --no-install-recommends postgresql-common postgresql-11 postgresql-client-11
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -37,6 +37,7 @@ jobs:
 
       - name: Install/Link Postgres
         run: |
+          sudo apt update
```
willronchetti, line 40: Alex added `sudo apt-get update -y -q` - not sure if the distinction matters here, but if the build ran/passed I'm sure it's fine.

willronchetti (approved):

.github/workflows/main.yml
netsettler: I added -yq in the style of the subsequent argument. (The online doc confirms those are both OK settings. I didn't have an interactive environment to try so we'll see when GA runs.)

PR #1436, opened 2021-02-18
New HiGlassEmbeddedInstance Placeholder

utku-ozturk:
**Trello:** https://trello.com/c/b2dv1OkQ

A simple placeholder was added to call HiGlass view config items by `uuid`. This new placeholder will make embedding multiple instances of HiGlass items into a single static section easy.

**Usage: (live demo [here](http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/static-sections/2df51046-530e-4737-af12-3ec62ed1870f/))**
```javascript
<p>Below are the insulation scores and boundary calls produced by different members of the 4DN Network along with the results produced by DCIC.</p>
<h3>Insulation scores for H1-hESC (Tier 1) Micro-C</h3>
<HiGlassEmbeddedInstance uuid="712164de-26e9-4faf-8446-174ba2758442" key="higlass-1" />
<h3>Insulation scores for H1-hESC (Tier 1) Hi-C </h3>
<HiGlassEmbeddedInstance uuid="9ebd05ce-7a68-48a8-90c9-9c22539772ff" key="higlass-2" />
<h3>Insulation scores for HFFc6 (Tier 1) Hi-C</h3>
<HiGlassEmbeddedInstance uuid="db8d5008-72ea-4ee7-9ca4-149ae9874a98" key="higlass-3" />
<h3>Insulation scores for HFFc6 (Tier 1) Hi-C</h3>
<HiGlassEmbeddedInstance uuid="10e3b0d2-c71c-4291-a894-6ae1933baaea" key="higlass-4" />
```


alexkb0009 (approved):
Looks/works good from what can tell.

Might also have been able to re-use HiGlassAjaxLoadContainer to re-use its state.loading + state.higlassItem functionality but doesn't matter much and unsure about if what other stuff that HiGlassAjaxLoadContainer does (scaleViewconfToHeight) is necessary.

#### Off-Topic / Possible Future To-Do

(maybe can add a backlog ticket for this)

When HiGlassAjaxLoadContainer was first made, it was assumed there'd only ever be 1 instance per view/page. In practice, it's apparent now that we might have multiple HiGlass instances per view/page. It would make sense to convert [this function](https://github.com/4dn-dcic/fourfront/blob/embedded_higlass/src/encoded/static/components/item-pages/components/HiGlass/HiGlassAjaxLoadContainer.js#L25) from being globally memoized to just being a static function and then memoize the function per instance in the constructor (i.e. `this.memoized = { scaleViewconfToHeight: memoize(HiGlassAjaxLoadContainer.scaleViewconfToHeight) }` in constructor and then use `this.memoized.scaleViewconfToHeight` in the render method instead of `HiGlassAjaxLoadContainer.scaleViewconfToHeight`) for a slight performance boost.

(An even better update might be to move the `scale1dTopTrack` functionality/func out of HiGlassAjaxLoadContainer completely and put it into a separate component to have better modularity/separation-of-concerns or move it into HiGlassPlainContainer at least, since the scale1dTopTrack functionality is very unrelated to the AJAX loading of the HiGlass Item.)

utku-ozturk:
@alexkb0009 Re: HiGlassAjaxLoadContainer

Yes, I definitely agree with your concern about HiGlassAjaxLoadContainer. Actually, the first implementation was calling  HiGlassAjaxLoadContainer directly. Then we noticed that BasicUserContentBody also have some nice features: description and displaying explore button on top-right side.

utku-ozturk:
@alexkb0009 Re: Off-Topic / Possible Future To-Do

Created a new [ticket](https://trello.com/c/WurRk8qd).

PR #1438, opened 2021-02-23
Port UI Security Updates to 4DN

utku-ozturk:
Trello: https://trello.com/c/b20SNu0D

CGAP security updates are ported into 4DN.


src/encoded/static/components/forms/UserRegistrationForm.js
```diff
@@ -183,9 +183,10 @@ export default class UserRegistrationForm extends React.PureComponent {
             ajax.load(
                 endpoint,
                 (resp) => {
+                    const jwtToken = resp.getHeader('X-Request-JWT');
                     // TODO
                     this.setState({ 'registrationStatus' : 'success-loading' });
-                    onComplete(); // <- Do request to login, then hide/unmount this component.
+                    onComplete(jwtToken); // <- Do request to login, then hide/unmount this component.
```
alexkb0009, line 189: This edit isn't necessary. `onComplete`/`onRegistrationComplete` is already bound to `this` & `idToken`, so don't need to pass anything else into it. [Checkout bind at MDN.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) (`bind` clones a function and sets the `this` of the newly-cloned function to first parameter of the `bind` call, and sets subsequent parameters of the `bind` call as params always passed into the cloned function).

utku-ozturk: Kk, reverted the commit.

gulpfile.js
```diff
@@ -84,20 +86,19 @@ function watch(){
...
+    const origPath = sharedComponentPath;
 
-    try { // Get exact path to dir, else leave. Used to avoid needing to webpack dependency itself.
-        for (var i = 0; i < 10; i++) { // Incase multiple links.
-            sharedComponentPath = fs.readlinkSync(sharedComponentPath);
-            isLinked = true;
-        }
-    } catch (e){
-        // ... not linked
-    }
```
alexkb0009, line 96: This existing code didn't work with `nvm` that well, it would grab the correct path of the symlink once, but then if that correct path was also a symlink, it wouldn't necessarily find the real path. Using `fs.realpathSync(sharedComponentPath);` appears to be more stable. If works well will migrate to CGAP.

alexkb0009 (approved):

src/encoded/authentication.py
```diff
@@ -279,19 +340,41 @@ def logout(context, request):
     The front-end handles logging out by discarding the locally-held JWT from
     browser cookies and re-requesting the current 4DN URL.
     """
-    #request.session.invalidate()
-    #request.response.headerlist.extend(forget(request))
+    request_parts = urlparse(request.referrer)
```
netsettler, line 343: Does request really spell `referrer` correctly? I thought it was lost in antiquity that we were doomed to see "referer" spelled wrong in all header situations. As long as the right thing is happening here, I'm fine. But my eyes are now trained to expect a misspelling.

src/encoded/authentication.py
```diff
@@ -279,19 +340,41 @@ def logout(context, request):
...
+
+    # Deletes the cookie
+    request.response.set_cookie(
+        name='jwtToken',
+        value=None,
+        domain=request_domain,
+        max_age=0,
+        path='/',
+        overwrite=True
+    )
```
netsettler, line 354: Does this not need to specify the http_only and secure options in order to be able to find the cookie it wants to delete? or does it not matter?

src/encoded/authentication.py
```diff
@@ -192,19 +210,22 @@ def get_user_info(request):
         request.set_property(get_user_info, "user_info", True)
         return email
 
-    def get_token_info(self, token, request):
+    @staticmethod
+    def get_token_info(token, request):
         '''
```
netsettler, line 215: Very minor: If you see any triple-single-quotes in these files and want to upgrade them triple-double-quotes, feel free. PEP8 standard is triple-double-quotes for doc strings.

src/encoded/authentication.py
```diff
@@ -228,41 +249,81 @@ def get_token_info(self, token, request):
...
+        try:
+            # Ensure this is a JWT token, not basic auth.
+            # Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication and
+            # https://tools.ietf.org/html/rfc6750, JWT is introduced by 'bearer', as in
+            #   Authorization: Bearer something.something.something
+            # rather than, for example, the 'basic' key information, which as discussed in
+            # https://tools.ietf.org/html/rfc7617 is base64 encoded and looks like:
+            #   Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==
+            # See also https://jwt.io/introduction/ for other info specific to JWT.
+            [ auth_type, auth_data ] = request.headers['Authorization'].strip().split(' ', 1)
```
netsettler, line 263: Not urgent, so doesn't have to be done for this PR, but: I'd like to see a separate subroutine for parsing out this stuff. I hate seeing `.strip()` and `.split(...)` used amidst higher level operations. This kind of formula is done differently in different places and is a risk of problem. Some sort of separately tested `parse_authorization` would be good.

src/encoded/authentication.py
```diff
@@ -279,19 +340,41 @@ def logout(context, request):
...
+        overwrite=True
+    )
+
+    request.response.status_code = 401
+    request.response.headers['WWW-Authenticate'] = (
+        "Bearer realm=\"{}\", title=\"Session Expired\"; Basic realm=\"{}\""
+        .format(request.domain, request.domain)
+    )
+
+    return { "deleted_cookie" : True }
```
netsettler, line 362: This is the result of the login endpoint?

netsettler (commented):
Not done with review yet but here are a few comments.

src/encoded/authentication.py
```diff
@@ -228,41 +249,81 @@ def get_token_info(self, token, request):
...
+    request.response.set_cookie(
+        "jwtToken",
+        value=request_token,
+        # THE BELOW NEEDS TESTING RE: CLOUD ENVIRONMENT:
+        domain=request_domain,
+        path="/",
+        httponly=True,
+        samesite="strict",
+        overwrite=True,
+        secure=is_https
```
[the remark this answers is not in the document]
alexkb0009: (Will respond on Slack)

src/encoded/authentication.py
alexkb0009: The /logout endpoint

src/encoded/authentication.py
alexkb0009: It shouldn't matter I don't think, at least testing showed it to work. As long as the request is still to an https endpoint, then it will unset the https.

One place it didn't work is because I set my localhost for CGAP to https, and 4DN to be http. Cookie was shared between both instances, but http://localhost couldn't delete a cookie created by https://localhost. Which I think is to be expected.

src/encoded/authentication.py
alexkb0009: Haha.. at least in [recent version it looks like it has both](https://gyazo.com/1f77af171b1bff3237af0343d757e82b).

PR #1441, opened 2021-02-28
Use elasticsearch 6.8.1 (C4-612)

netsettler:
This ports a subset of changes from my other PR in order to get things to a place where master is working correctly:

* I ported a number of changes to `Makefile`, some that are needed for this to work, and some that are just incidental but already debugged where it was easier to just take what I was already testing on another branch. The important bit is to separate `make test` into two parts, where the workbook tests are separated from the non-workbook tests. That keeps the tests that are order-dependent from hurting each other. The actual changes are:

  * `make test` now calls `make test-unit` and `make test-npm`. The names are for consistency with `cgap-portal` and in fact there is no `npm` testing going on here. The real separation is, as noted above, between indexing and non-indexing tests. We'll look at renaming this later.
  * `make travis-test` is a tag added just for completeness. This is all about GitHub Actions now, not Travis. But on GA, the two aprts of `travis-test` are invoked separately. The `travis-test` target is more for completeness to s how how these relate as part of a whole.

* A number of PyTest "marks" were added. They have documentation, so I won't bother to describe them here. The main thing I'll say here is that `action_fail` is gone and everything consistently just uses `broken` now. There are some newly added ways to keep something from running, see the `pytest.ini` for an explanation of why that might be. In a later PR, ways of invoking performance and integration tests will follow.

* Several tests that were marked `action_fail` are now more simply marked as `broken`. And the `make` documentation output was updated a bit to match.


willronchetti (approved):

Makefile
```diff
@@ -38,7 +40,7 @@ macpoetry-install:  # Same as 'poetry install' except that on OSX Catalina, an e
 
 configure:  # does any pre-requisite installs
 	pip install --upgrade pip
-	pip install poetry==1.0.10  # pinned to avoid build problems we cannot fix in pyproject.toml
+	pip install poetry==1.1.4  # poetry latest as of 1/25/2021 seemed to work but apparantly does not
```
willronchetti, line 43: Comment is out of date, this version is also misaligned with deployment (1.1.0).

willronchetti (approved):

netsettler:
I've done a test deploy on fourfront-mastertest with the most recent version of this PR and it's deploying OK. Trying to be very careful about details.

netsettler:
Actually, I changed one last comment and pushed that up but I won't wait for a test on that. If that breaks the build, we are in trouble. Going to merge now. It will re-test on merge anyway.

PR #1445, opened 2021-03-05
Sequential imaging

andreacosolo:
Schema changes to accommodate multiplexed, multimodal imaging datasets.
- new properties in imaging_path and experiment_mic
- imaging_path fields embedded in experiment
- inserts
- [unrelated] label is now a suggested enum


src/encoded/schemas/experiment_mic.json
```diff
@@ -84,7 +84,7 @@
                     },
                     "channel":{
                         "title": "Imaging channnel",
-                        "description" : "channel info, ie. ch01, ch02...",
+                        "description" : "channel info, ie. ch00, ch01...",
```
aschroed, line 87: With the changes and what you have said about there really only being a max of 4 channels used - should this be restricted to an enum or suggested_enum - or do you think people will go ahead and build >4 channel scopes just to spite us if we did?

src/encoded/schemas/experiment_mic.json
```diff
@@ -118,6 +118,28 @@
...
+                "fluorophore cleavage",
+                "photobleaching",
+                "strand displacement"
+            ]
+        },
+        "fiducial_label": {
+            "title": "Fiducial Label",
+            "description": "Type of fiducial label used for drift correction during sequential imaging",
+            "type": "string",
+            "lookup": 69,
```
aschroed, line 138: Not that we've actually implemented it in a rigorous way but if these fields are specific to sequential imaging techniques as indicated by the descriptions would it make sense to add a "use_for": "sequential imaging" similar to how we do for some fields in for example ExperimentSeq?

aschroed (approved):
I think this looks good.  I've left a couple of non-blocking comments. 

Regarding your points about being able to search and display all the relevant info - for different types of "sub" assays for each sequential imaging experiment it is a good point and I'm not sure as far as searching how important it will be but it will certainly be useful to have this information displayed on the experiment or experiment set page (and I guess if that is the case it should be searchable).  For now I think we should go with what you have, get some real experimental metadata in and then we can get a better idea of how to display and enhance search capability.

src/encoded/schemas/experiment_mic.json
andreacosolo: It's actually up to 5 channels and it makes sense to have this as enum. If there will ever be more than 5 channels, then more things need to be changed (i.e. MicroscopeSettings), so at that point we can revise also this enum.

src/encoded/schemas/experiment_mic.json
andreacosolo: Sure, I didn't know about this. How is it used? By submit4dn or FFedit-create?

PR #1448, opened 2021-03-10
Demo/Help Pages for HiGlassEmbeddedInstance and ExperimentSetMatrix

utku-ozturk:
Trello: https://trello.com/c/1utYPJAQ


src/encoded/static/data/experiment-set-matrix-demo.jsx
```diff
@@ -0,0 +1,235 @@
...
+    <p>This generates the following display:</p>
+    <ExperimentSetMatrix
```
aschroed, line 18: I think it would be really helpful if the actual example matrix was cleaned up a bit so extra unnecessary stuff was not included - or if it does need to be there some textual explanation as to why would be good.

I notice that for example, the search returns results for K562 cells and these are included in the matrix but K562 doesn't look like it's included in the headerColumnsOrder property.

Also you had "experiment_category": "experiments_in_set.experiment_type.experiment_type", as the first entry in the FieldChangeMap (which I changed to experiments_in_set.experiment_type.experiment_subclass_short like it is in the following example) but that doesn't seem to be used? So my change had no effect.  

While the examples are indeed very helpful it's just a bit hard to tell what is actually being used and needed to get the matrix built.

Also it looks like clicking on a box of the martrix currently is not returning the search results in the correct numbers - I don't know if this is just because it's searching data but deployed on hotseat or local?

aschroed (commented):
Hi Utku - I've started looking at the ExperimentSetMatrix demo and left a longer comment there.
I will try to get to the embedded higlass display soon.  The demos are very useful but I think a little clean up of the examples could make them much clearer.

src/encoded/static/data/experiment-set-matrix-demo.jsx
utku-ozturk: @aschroed Thanks for your valuable feedback. I'll add more detailed explanations for the props shortly. For first-time users, it doesn't look easy to use.

I write down a short comment below regarding your feedbacks:

> I notice that for example, the search returns results for K562 cells and these are included in the matrix but K562 doesn't look like it's included in the headerColumnsOrder property.

> Also it looks like clicking on a box of the martrix currently is not returning the search results in the correct numbers - I don't know if this is just because it's searching data but deployed on hotseat or local?

For the sake of example, queries."4DN".url explicitly starts with `"https://data.4dnucleome.org/browse/?..."` since `mastertest` returns no result in this case. In a real-world example, starting with`"/browse/?..."` is recommended. On the other hand, querying `data` from `mastertest` not returns the exact counts since we run a cross-site query. (e.g. [the query](https://data.4dnucleome.org/browse/?award.center_title=NOFIC+-+Belmont&experiments_in_set.biosample.biosource_summary=K562+%28Tier+2%29&experiments_in_set.biosample.biosource_summary=HCT116+%28Tier+2%29&experiments_in_set.biosample.biosource_summary=HFFc6+%28Tier+1%29&experiments_in_set.biosample.biosource_summary=H1-hESC+%28Tier+1%29&experiments_in_set.experiment_type.display_title=TSA-seq&experiments_in_set.experiment_type.display_title=DamID-seq&experiments_in_set.experiment_type.display_title=2-stage+Repli-seq&status=released&status=released+to+project&limit=all) returns 83 items for not logged-in, but 97 for authenticated users)

aschroed:
Hi Utku.  I had a look at the embedded higlass view demo page and this looks good - although nothing showed up for the examples.  I guess maybe these are place holders?  Or are these test views that are not available on local?  In any case aside from this point I think this page is good.

utku-ozturk:
@aschroed Re: nothing showed up for the examples

I think something went wrong since icons are not just placeholders; they're preloaders of actual HiGlass view configs.  I pulled the recent commits and re-ran 4DN locally. HiGlassEmbeddedInstance demo page looks:

Firefox_Screenshot_2021-03-15T15-40-27 523Z

alexkb0009 (approved):
Looks fine from what can tell.

utku-ozturk:
@aschroed We completely rewrote the `ExperimentSetMatrix` demo page and deployed it into mastertest: http://mastertest.4dnucleome.org/static-sections/eef41c6d-a8c9-4e4e-8a8e-e44df2ad3dc3/. As you may remember, even if you run the demo locally, the data was gathered from https://data.4dnucleome.org and the count in the boxes did not match the counts when it was clicked. So, we reimplemented it, and now the demo (except the last example) relies on the data supplied by inserts only.

aschroed (approved):

PR #1449, opened 2021-03-10
validation_errors Facet Removed

utku-ozturk:
As the title indicates, the `validation_errors` facet is removed. (This has been a topic that has been discussed for a long time, finally fixed in the simplest way as discussed in UI team meeting.)


src/encoded/tests/test_indexing.py
```diff
@@ -232,6 +232,7 @@ def test_file_processed_detailed(app, testapp, indexer_testapp, award, lab, file
     assert found_rel_sid > found_fp_sid  # sid of related file is greater
 
 
+@pytest.mark.skip(reason="validation_errors facet removed in search.py")
```
willronchetti, line 235: I think this test should still work... Does it not?

src/encoded/tests/test_validation_errors.py
```diff
@@ -7,11 +7,17 @@
...
-    pytest.mark.working, pytest.mark.indexing,
-    # pytest.mark.flaky(rerun_filter=delay_rerun),
-]
-
+# skip all tests tests since validation_errors removed in search.py
+# https://stackoverflow.com/a/42512169
+if pytest.__version__ < "3.0.0":
+  pytest.skip()
+else:
+  pytestmark = pytest.mark.skip
```
willronchetti, line 15: Are you running Python < 3.6 anywhere? I don't think is needed. We don't officially support <3.6 so I am hesitant to accept this change.

src/encoded/tests/test_indexing.py
utku-ozturk: @willronchetti My bad, this test is not related to facets. I'll remove the decorator shortly.

src/encoded/tests/test_validation_errors.py
utku-ozturk: @willronchetti Well, you are right; let me recheck it. (At first, I had added `pytest.mark.skip` decorator to tests in test_validation_errors.py to skip all tests instead of deleting the file, but both local and Github Actions failed. At the end of some trials and errors, those lines above help me skip the tests.)

willronchetti (approved):

PR #1450, opened 2021-03-10
HiGlass Container Performance Updates

utku-ozturk:
Trello: https://trello.com/c/WurRk8qd


alexkb0009 (approved):
Makes sense / looks good!
Good to merge assuming working as expected.

PR #1460, opened 2021-03-17
Open Data Downloads, C4-590

ebberg:
This sets the location for the `@@download` link redirects to one of two possible locations:
1) download redirect via a presigned url to a 4DN S3 bucket
2) download redirect to the public S3 URL of data released to the Open Data S3 bucket

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -726,8 +726,9 @@ class File(Item):
     def get_open_data_url_or_presigned_url_location(self, external, request, filename) -> str:
         """  Returns the Open Data S3 url for the file if present (as a calculated property), and otherwise returns
             a presigned S3 URL to a 4DN bucket. """
-        if self.open_data_url:  #
-            return self.open_data_url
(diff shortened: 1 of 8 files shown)


src/encoded/types/file.py
```diff
@@ -706,6 +706,89 @@ def extra_files_creds(self):
...
+            'Bucket': external['bucket'],
+            'Key': external['key'],
+            'ResponseContentDisposition': "attachment; filename=" + filename
+        }
+        if request.range:
+            param_get_object.update({'Range': request.headers.get('Range')})
+        location = conn.generate_presigned_url(
+            ClientMethod='get_object',
+            Params=param_get_object,
+            ExpiresIn=36*60*60
```
netsettler, line 722: Very small point but if you 
```
from dcicutils.misc_utils import as_seconds
```
you can use `as_seconds(hours=36)` here and it might be a little clearer.

aschroed (approved):
This looks good to me.

PR #1462, opened 2021-03-23
Change to track_and_facet_info calculated property

aschroed:
UPDATE - decided to change the way regular attribution was added to processed files so the calcprop has been simplified 

~~- Add logic to make 4DN-DCIC the generating_lab if a file is derived from a 4DN pipeline run.  Otherwise the generating_lab is the lab that submits the file i.e. File.lab~~
- Add track_and_facet_info.experimental_lab to be the lab attribution of the experiment or replicate set to which the file in linked - won't exist if the file is not linked.
- Added tests.
- Updated columns and facets in schemas.


src/encoded/tests/test_file.py
```diff
@@ -1240,10 +1241,128 @@ def test_track_and_file_facet_info_file_patch_override_fields(
...
+        testapp, proc_file_json, experiment_data, lab, dcic_lab):
+    """ case where experiment is from a lab and file generated by
+        dcic pipe - dcic is made contributing lab for file - encode is lab
+    """
+    experiment_data['lab'] = proc_file_json['lab'] = lab.get('uuid')
+    proc_file_json['contributing_labs'] = [dcic_lab.get('uuid')]
+    pf = testapp.post_json('/file_processed', proc_file_json).json['@graph'][0]
+    experiment_data['processed_files'] = [pf['@id']]
+    testapp.post_json('/experiment_hi_c', experiment_data, status=201)
+    pf2 = testapp.get(pf['@id']).json
```
willronchetti, line 1266: Not clear to me why these GETs are required here... @@object is returned from the post above so you already have `track_and_facet_info` computed from the original post `pf`? Same question throughout these tests.

I suspect the act of posting the experiment changes the calc prop? In that case what you're doing here works by chance because of the mapping bug we discussed previously in this testing (these tests use default mappings). This makes ES item unique key resolution fail, so instead of getting the cached value of `track_and_facet_info` from ES it is recomputed from the DB and gives (presumably) the correct result below. 

My point is if you ran these tests in a production environment they would probably fail because the GET would be correctly handled by ES and you'd get a cached value for the calc prop. You would need to add a sleep after posting the experiment so reindexing of fp can occur.

willronchetti (approved):
These changes generally look good just had a clarifying question on the tests. We should also note that this change will slow the indexing/retrieval process of this item type as it includes additional GETs.

src/encoded/tests/test_file.py
```diff
@@ -1240,10 +1241,128 @@ def test_track_and_file_facet_info_file_patch_override_fields(
...
+        'awards': [award['@id']],
+        'uuid': '9780e6b5-507c-4287-b817-103e784587c8'
+    }
+    return testapp.post_json('/lab', item).json['@graph'][0]
+
+
+def test_track_and_file_facet_info_pf_attr_2_dcic_w_expt_lab(
+        testapp, proc_file_json, experiment_data, lab, dcic_lab):
+    """ case where experiment is from a lab and file generated by
+        dcic pipe - dcic is made contributing lab for file - encode is lab
```
andreacosolo, line 1259:
```suggestion
        dcic pipe - dcic is made lab_name for file - encode is experimental_lab
```

andreacosolo (approved):
Just a couple of minor comments/edits

src/encoded/tests/test_file.py
aschroed: Interesting - to answer your question - yes posting the experiment with the file linked to it changes the value of the calcprop on the file.  I wonder of the bug you found was actually there by design just for this reason?  These are certainly not the only tests that depend on rapid sequential posts/patches upon which expected changes are then asserted likely sooner than indexing could occur.

PR #1464, opened 2021-03-25
Small download fixes

willronchetti:
- Allow `open_data_url` to compute `filename` from `accession` and `file_format` so that it will resolve for all files
- Add fallback to DB in case calc prop is not present (could be removed later)
- Refactor GA into helper function, cleanup download 
- Remove custom range query handling, forward in header instead


aschroed (approved):
Looks like this is the better way to do this as this seems to match what is used in building (and updating) credentials once files are on s3 so will avoid any issues with odd filename values - not that that should happen at all but you never know.

PR #1465, opened 2021-03-25
A separate attempt at the renderers fix, and some related changes (C4-627)

netsettler:
* In `Makefile`
  * Implement `psql-test`.
  * Implement `kibana-start-test`.
  * Port `renderers.py` from `cgap-portal` except for small differences needed to preserve existing behavior.
  * Port `test_renderers.py`, too.
* Move `INDEXER_NAMESPACE_FOR_TESTING` setup from `conftest.py` to `conftest_settings.py`.
* Move `ORDER` from `datafixtures.py` to `conftest_settings.py`.
* Fix static help page tests like they're fixed in `cgap-portal` (involving changes to `page.json` and `static_section.json` in `workbook-inserts`.
* Mostly PEP8 fixes (a lot of them) in `test_indexing.py`, but also a minor change in how a mock is set up (via `mock.patch.object` instead of `mock.patch`).
* PEP8 changes (a lot) in `test_search.py`
* Make `workbook_fixtures.py` use `INDEXER_NAMESPACE_FOR_TESTING`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,10 +95,16 @@ deploy2:  # spins up waittress to serve the application
 	pserve development.ini
 
 psql-dev:  # starts psql with the url after 'sqlalchemy.url =' in development.ini
-	@psql `grep 'sqlalchemy[.]url =' development.ini | sed -E 's/^.* = (.*)/\1/'`
+	@scripts/psql-start dev
 
-kibana-start:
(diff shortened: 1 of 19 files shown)


willronchetti (approved):
LGTM!

PR #1469, opened 2021-04-04
reorder primary and secondary probes in ImagingPath

andreacosolo:
Re-ordering the `display_title` of ImagingPath so that, if `target` (e.g. antigen), `other_probes` (e.g. primary antibody) and `labeled_probe` (e.g. secondary antibody) are all present, they appear in a logic order.
Updating tests accordingly.


aschroed (approved):
Looks OK to me.

PR #1470, opened 2021-04-06
add antibodies in imaging path

andreacosolo:
- added antibodies to imaging_path as linkTo Antibody items
- in `display_title`, primary Ab and secondary Ab are equivalent to other_probes and labeled_probe, respectively
- display_title removes a label if already present in labeled probe
- display_title can be overridden
- embedded all fields from linkTo items used in their display titles
- since the ImagingPath display_title changed, modified the embedding list of all items that linkTo ImagingPath


src/encoded/types/imaging.py
```diff
@@ -40,22 +47,41 @@ def display_title(self, request, target=None, labeled_probe=None, other_probes=N
...
+        if labeled_probes:
+            labeled_probes = ", ".join(labeled_probes)
+
+        primary_probes = []
         if other_probes:
-            other_probes = ", ".join(other_probes)
+            primary_probes.extend(other_probes)
+        if primary_antibodies:
+            for ab in primary_antibodies:
+                antibody = request.embed(ab, '@@object').get('antibody_name')
```
aschroed, line 64: another use of request.embed

src/encoded/types/imaging.py
```diff
@@ -40,22 +47,41 @@ def display_title(self, request, target=None, labeled_probe=None, other_probes=N
                 target_title = request.embed(a_target, '@@object').get('display_title')
```
aschroed, line 47: Can you change from using request.embed  to get_item_or_none here (which I guess was pre-existing) and the following 2 instances - while they are really essentially the same the preferred usage is get_item_or_none - mainly I think due to better error handling?

aschroed (approved):
A couple of non blocking comments.  I presume you included enough test cases to ensure yourself that most if not all potential edge cases will be handled correctly.  Look good to me!

src/encoded/types/imaging.py
andreacosolo: I just used what was there before, didn't know that this was the preferred usage. Changed now.

aschroed (approved):
This looks good to me.  You may want Will to have a quick look as the helper functions and dependencies to create the embedded lists are new but they look good to me.

willronchetti (approved):

PR #1471, opened 2021-04-07
Column Renderer Func. Updates for track_and_facet_info.experimental_lab and track_and_facet_info.lab_name

utku-ozturk:
Trello: https://trello.com/c/Ju3jVRe5

Screen Shot 2021-04-08 at 13 31 02


src/encoded/static/components/browse/columnExtensionMap.js
```diff
@@ -151,10 +151,11 @@ export const columnExtensionMap = _.extend({}, basicColumnExtensionMap, {
...
                 track_and_facet_info: { lab_name } = {},
-                lab : { display_title: labTitle } = {}
+                lab : { display_title: labTitle } = {},
+                contributing_labs
             } = result;
             if (!lab_name) return null;
-            if (lab_name && labTitle && lab_name === labTitle){
+            if ((lab_name && labTitle && lab_name === labTitle) || (contributing_labs && _.contains(_.pluck(contributing_labs, 'display_title'), labTitle))) {
                 // If same exact lab name as our lab.display_title, then we just use lab render method to get link to lab.
                 return labDisplayTitleRenderFxn(...arguments);
```
alexkb0009, line 160: This [`labDisplayTitleRenderFxn`](https://github.com/4dn-dcic/fourfront/blob/b820e43c5465dcb83555504a8ab0cf284620d682/src/encoded/static/components/browse/columnExtensionMap.js#L17) will only render `result.lab` at moment it seems, even if it's the contributing lab. Is this intended functionality? If not, would probably change to be ~:

```javascript
const {
    track_and_facet_info: { lab_name } = {},
    lab : { display_title: labTitle } = {},
    contributing_labs = []
} = result;

if (!lab_name) return null;
if (labTitle && lab_name === labTitle){
    // If same exact lab name as our lab.display_title, then we just use lab render method to get link to lab.
    return labDisplayTitleRenderFxn(...arguments);
}
const contributingLabFound = contributing_labs.find(function({ display_title: contributingLabTitle }){
    return contributingLabTitle === lab_name;
});
if (contributingLabFound){
    // A little hacky but works.
    return labDisplayTitleRenderFxn({ ...result, "lab": contributingLabFound }, props);
}

return <span className="value">{ lab_name }</span>;
```

To be a bit cleaner, could create `<LabColumn lab={..} />` component and re-use it here instead of labDisplayTitleRenderFxn. And then either have labDisplayTitleRenderFxn render out a LabColumn also (or remove labDisplayTitleRenderFxn from code completely and replace usage of it w. LabColumn)

src/encoded/static/components/browse/columnExtensionMap.js
```diff
@@ -55,7 +55,7 @@ export const columnExtensionMap = _.extend({}, basicColumnExtensionMap, {
     'lab.display_title' : {
         'title' : "Lab",
         'widthMap' : { 'lg' : 200, 'md' : 180, 'sm' : 160 },
-        'render' : labDisplayTitleRenderFxn
+        'render' : LabColumn
```
alexkb0009, line 58: I'm not sure this would work ^ (pls double-check), I think may need to do ~:

```javascript
'render' : function(result, props){
    const { lab, submitted_by : { display_title : submitterTitle } = {} } = result;
    if (lab) {
        return <LabColumn {...{ lab, submitterTitle } />;
    }
    return null;
}
```

utku-ozturk: @alexkb0009  Uğur updated the render function.

alexkb0009 (approved):

PR #1473, opened 2021-04-12
Micrometa Path Updates

utku-ozturk:
Updated relevant file paths due to the updates in schema, image, and dimension file paths in MicroMeta repos.


ugur-camoglu (approved):
Looks good!

PR #1475, opened 2021-04-15
Small download adjustment

willronchetti:
- Allow open data URLs for archived files
- Fix extra_file download resolution when using cached value (trigger recompute with correct filename)

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -823,6 +823,8 @@ class File(Item):
             if hasattr(self.model, 'source'):
                 es_model_props = self.model.source['embedded']
                 open_data_url = es_model_props.get('open_data_url', None)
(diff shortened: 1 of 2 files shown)


src/encoded/types/file.py
```diff
@@ -823,6 +823,8 @@ def get_open_data_url_or_presigned_url_location(self, external, request, filenam
             if hasattr(self.model, 'source'):
                 es_model_props = self.model.source['embedded']
                 open_data_url = es_model_props.get('open_data_url', None)
+                if filename not in open_data_url:  # we requested an extra_file, so recompute with correct filename
```
aschroed, line 826: I guess there is guaranteed to be a filename?  If not an exception could be thrown.

aschroed (approved):
Looks good - one small comment as I wasn't able to completely verify that one value would always be present.

PR #1479, opened 2021-04-21
bug fix for some facets

andreacosolo:


src/encoded/schemas/antibody.json
```diff
@@ -80,13 +80,19 @@
...
-            "title" : "Vendor"
+        "antibody_vendor.display_title" : {
+            "title": "Vendor"
+        },
+        "antibody_target.display_title": {
+            "title": "Target"
         }
     },
     "columns": {
         "antibody_vendor": {
```
aschroed, line 91: Is not a similar change required here?

aschroed (approved):
Have these never worked as expected or did something change due to embedding changes?

Also one in line comment.

src/encoded/schemas/antibody.json
andreacosolo: Surprisingly it works without that change (not sure why though).

andreacosolo: I suspect these have never worked, but it's not something used too often to really notice.

aschroed: I had thought that the default when a property was not included was to use display_title and apparently that works for columns but I guess not for facets.

PR #1481, opened 2021-04-22
Fix treatment embeds

andreacosolo:


aschroed (commented):
I think it might be better to add the treatment_type to the dependencies.py mapping as it looks like it will be used everywhere that treatment gets embedded once the embedding method gets implemented for other item types eg. Experiment and ExperimentSet.  What do you think?

andreacosolo:
I completely agree. `dependencies.py` should have all embeds necessary for the `display_title` of Treatment items, and indeed many are missing (including `treatment_type`).

aschroed (approved):
I don't think there is much if any overhead in including Item-specific fields in the embedded list of the parent abstract Item but I am really not sure if this would cause any issues at all so this is a question for Will.  Depending on his answer this looks good to me.

willronchetti (approved):

PR #1482, opened 2021-04-23
Adjust library dependencies for some unused and less-used libraries (C4-653)

netsettler:
This implements some changes called for in [C4-653](https://hms-dbmi.atlassian.net/browse/C4-653):

* Remove `awscli` dependency as not needed.
* Move `docutils` and `loremipsum` to dev dependency status.
* Remove `rfc3987` dependency in favor of functionality from `dcicutils.misc_utils.is_valid_absolute_uri`.

This also opportunistically adds these libraries as dev dependencies:

* `pipdeptree`
* `pip-licenses`


pyproject.toml
```diff
@@ -120,10 +116,12 @@ boto3-stubs = ">=1.17.15.0"
 coverage = ">=5.3.1"
 codacy-coverage = ">=1.3.11"
 coveralls = ">=3.0.0"
+docutils = "0.12"
 flake8 = ">=3.8.4"
 flaky = ">=3.7.0"
 # flask only for moto[server]
 flask = ">=1.1.1"
+loremipsum = "1.0.5"
```
willronchetti, line 124: Its not clear to me why moving these to dev-dependencies helps us - how is it that it isn't an issue here? Because it's considered an "add-on"?

willronchetti (approved):

pyproject.toml
netsettler: It may not help at all. It will make it clearer if we end up doing the library form of this that I've talked about doing it. But for now it's a symbolic indication they're not needed at runtime. The difference from `cgap-portal` is that it didn't have uses of these at all, while `fourfront` has some utility scripts that do use them.

PR #1487, opened 2021-04-29
dependencies.py refactor + add vendor to Antibody display_title

andreacosolo:
This makes it more complete and citation-like. Example: anti-tubulin (Abcam, ab1234).

- update Antibody display_title
- update tests
- add antibody to dependencies.py
- update all items to embed default dependencies for antibody
- slight refactor of dependencies.py to reduce redundancy in EMBED_MAPPER lists
- updated a test to reflect a new embed that had been added to ontology_term for generate_ontology
- tweaked imaging_path and experiment_mic embedded list builds to use these updates


aschroed (approved):

willronchetti (approved):
I like this change a lot. Should probably be ported to CGAP at some point, though CGAP uses this feature less at this time so probably not a big priority.

PR #1488, opened 2021-05-06
ExperimentSetMatrix Sorting

utku-ozturk:
Trello: https://trello.com/c/jIa93meK

### Screencasts: 

Sort by counts: https://i.gyazo.com/990c9753f6d063aef338a7acfae3d6c5.gif
Sort by labels: https://i.gyazo.com/b3a14a09077ba5fa7414233af9801b8f.gif


src/encoded/static/components/static-pages/components/StackedBlockVisual.js
```diff
@@ -526,11 +583,43 @@ export class StackedBlockGroupedRow extends React.PureComponent {
         });
     }
 
+    //The function of creating sorting according to the total values of the data
+    //As a result, objects in the array are created. (groupingKey, count)
+    //For example ('groupingKey' :dataBinding, 'count': 146)
+    static sortBlock = memoize(function (data, groupedDataIndices, groupingKey, sortField) {
```
alexkb0009, line 589: If we have multiple matrices (or StackedBlockGroupedRow) on a single page/view, then would make sense to have this static sortBlock function not be memoized here (as static funcs are global and would be shared between all the matrix instances, negating the effect if more than 1 instance of the matrix is present), but instead to have per-instance memoization like `this.memoized = { sortBlock: memoize(StackedBlockGroupedRow. sortBlock) }`.

alexkb0009 (approved):
Looks great, left 1 comment re: where to memoize sortBlock potentially, but wouldn't prevent a merge if working otherwise and we need to get in ASAP.

src/encoded/static/components/static-pages/components/StackedBlockVisual.js
utku-ozturk: Thanks for the feedback. I moved memoization into the constructor. (https://github.com/4dn-dcic/fourfront/pull/1488/commits/227347d2d4c6f69659386c4e50c8377f2219a2a4)

PR #1491, opened 2021-05-18
Matrices qc

Lmercadom:


src/encoded/types/quality_metric.py
```diff
@@ -561,3 +561,31 @@ def round2(numVal):
...
+    schema = load_schema('encoded:schemas/quality_metric_mcool.json')
+    embedded_list = QualityMetric.embedded_list
+
+    @calculated_property(schema=QC_SUMMARY_SCHEMA)
+    def quality_metric_summary(self, request):
+        qc = self.properties
+        qc_summary = []
+        val = ''
+        qc_val = qc.get("Failed Balancing")
+        val = ';'.join(qc_val)
```
aschroed, line 585: maybe add a space after the semicolon '; '

.ipynb_checkpoints/Testing_Locally-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
```
aschroed, line 1: Need to remove this from branch.

Testing_Locally.ipynb
```diff
@@ -0,0 +1,61 @@
+{
```
aschroed, line 1: And remove this.

aschroed (changes requested):
Looking at your examples - generally the schema updates look fine, however, when the QC gets posted if should include the overall_quality_status value - the default in the generic QCMetric is 'PASS' but you should have a 'WARN' if some of the resolutions don't converge and probably 'FAIL' if none of them do.  Maybe you already have this bit in your foursight or workflow code but this QCMetricMcool example has 'PASS' when it should not.  And we should probably have a UI change to possibly show the overall status somewhere?

You also committed some of your jupyter notebook code?

And a small change to add a space after the semi-colon where you join the values.

aschroed (approved):

aschroed:
Looks good as long as tests pass I think this should be good to go.

PR #1492, opened 2021-05-18
Download Edits

utku-ozturk:
Trello: https://trello.com/c/KXRBnuHW

1. Download button's text for Xls/Xlsx file was `"View"`. It  is replaced with `"Download"`. ([SPC PR](https://github.com/4dn-dcic/shared-portal-components/pull/69))
2. If the attached file name has non-ASCII characters then download end-point throws `InvalidArgumentHeader value cannot be represented using ISO-8859-1.response-content-dispositioninline; filename=`. The file name is URL encoded to fix the issue.
3. Protocols search view has a Files column that has a downloadable/viewable file links. But pdf-like viewable files were displayed within the tab itself if clicked on them. Instead, it is now opened in a new tab to preserve context.
Screen Shot 2021-05-18 at 11 42 58
4. Image item page is added to display images. (Currently, details view is displayed  and it has no explicit view/download options) Moreover, if Image is related to Microscopy File, it grabs the thumbnail from the mic file and also display a link to Microscopy File item page.
Screen Shot 2021-05-18 at 12 18 12


aschroed (approved):
This all looks good from my perspective as a viewer/user.  

It's too bad that we only have thumbnails for microscopy files - adding the ability to see a thumbnail view of an image file on the overview section before one clicks on the 'View' button.  If you think this might be doable via the UI side - similar to how we can see an image in a Biosample Overview if present: https://data.4dnucleome.org/biosamples/4DNBSA2E69FQ/ it would be great to add a ticket for this.

utku-ozturk:
@aschroed Yes, we can add a thumbnail image into the Image item page as it is on the biosample overview. However, we must consider the following cons:

[Bisample Overview](https://data.4dnucleome.org/biosamples/4DNBSA2E69FQ/), actually displays the original image as a thumbnail. (it is ~8 MB and 2560x1920px - far above what should be and takes a noticeable time to download) `Image` item type also has no thumbnail field on its own unless it has a `microscopy_file` with a thumbnail. If we display the original image as a thumbnail, we should be ready for a similar lag while downloading.

alexkb0009 (approved):

PR #1493, opened 2021-05-19
File Item Page Updates

utku-ozturk:
Trello: https://trello.com/c/IOz7pMiU

Before merging this branch, pls check the SPC PR: https://github.com/4dn-dcic/shared-portal-components/pull/70

1. File's `track_and_facet_info.experiment_bucket` field is used to determine whether the file is supplementary. (`track_and_facet_info.experiment_bucket` should exist and be non-empty and also be neither '`processed file`' nor  '`raw file`'. We grabbed the logic from the calculated property's method attached below. Sounds good to us, but not sure 100%) If it is supplementary, append Supplementary text to classification. Otherwise, leave the classification as is.

https://github.com/4dn-dcic/fourfront/blob/23237ca51527abdb1690633f8103bcfd562bbb98/src/encoded/types/file.py#L397-L410

Screen Shot 2021-05-19 at 11 22 53

2. File Item Page lists experiment sets result in the `Overview` tab. When the user clicks on the result, the page is always navigated to the "`Experiment Set`" page with the default tab (1st tab) open. On the other hand, if the file is supplementary, navigating to the `Supplementary Files` tab would make the relationship more clear. (this is also true for raw and processed (but not supplementary) files) So, new targetTabKey prop (string) is added to embedded search tables. ([For more  info](https://github.com/4dn-dcic/shared-portal-components/pull/70))

https://i.gyazo.com/7f6f9d0146927bcffccb11a2423ee315.gif
https://i.gyazo.com/1347a4ac5d2169dc7536b05f1f93f654.gif

### DIFF
--- a/.ipynb_checkpoints/Testing_Locally-checkpoint.ipynb
+++ b/.ipynb_checkpoints/Testing_Locally-checkpoint.ipynb
@@ -1,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -3694,8 +3694,8 @@
       }
     },
     "@hms-dbmi-bgm/shared-portal-components": {
-      "version": "github:4dn-dcic/shared-portal-components#14cddcf51fc68f1aad4d7bee207c69fb9a67ebf4",
-      "from": "github:4dn-dcic/shared-portal-components#0.1.8b2",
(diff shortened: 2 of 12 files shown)


andreacosolo (approved):
Looks great to me! I did not review the code in detail, but I tested it on mastertest and I think now the relation between a supplementary file and the Experiment Set is more clear. Also the logic to identify a file belonging to opf sounds good to me.

alexkb0009 (approved):

PR #1496, opened 2021-05-27
Sum targets when imaging paths are split

andreacosolo:
Some ExperimentMic can have multiple imaging_paths with the same target. For example, 90 TADs can be imaged splitting '50 TADs' in one path and '40 TADs' in another. In this case, the Target listed by experiment_categorizer will detect them and sum them.

- modified experiment_categorizer for ExperimentMic
- added test

### DIFF
--- a/src/encoded/types/experiment.py
+++ b/src/encoded/types/experiment.py
@@ -1,6 +1,7 @@
 """Abstract collection for experiment and integration of all experiment types."""
 
 import itertools
+import re
 
 from snovault import (
     abstract_collection,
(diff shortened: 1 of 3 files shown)


src/encoded/types/experiment.py
```diff
@@ -1014,12 +1015,25 @@ def experiment_categorizer(self, request, experiment_type, biosample, imaging_pa
...
+                    summ = get_item_or_none(request, target, 'bio_feature')['display_title']
                     path_targets.append(summ)
             if path_targets:
-                value = ', '.join(list(set(path_targets)))
+                value = []
+                sum_targets = {}
+                for target in path_targets:
+                    # check if target starts with numbers, e.g. '50 TADs', '40 TADs'
+                    # sum them if there are more: '90 TADs'
+                    split_target = re.split(r'(^[0-9]+)([^0-9].*)', target)
```
aschroed, line 1028: Are you sure about this re.split?  It looks a bit tricky.  I guess it works as expected in python 3.6 but I think the second paren capture statement will return an empty string in v3.7 and above - so it may possibly not work exactly as expected in the future ?  Also is the not digit at the start of the second capture required?  Maybe a greedy ? at the end of the first capture would be safer?  And it seems we will need to take care to make sure that targets use the format expected - can we document this somewhere if that is the case?

andreacosolo: Right, the second capture group was not necessary. It should be fixed now.
The use case here is when target(s) of ImagingPath(s) start with numbers, and the rest of the text is identical. Very likely, this is only obtained by using a `preferred_label` for the BioFeature, which is already an advanced use case (typically, wranglers take care of it). I agree that documenting is necessary, although nothing happens in case the text is not formatted properly: targets will just remain split, as it has been until now.
The test should help understanding how it works, and now I also added docstrings. Any suggestion on how to document better how to use this?

aschroed (approved):

PR #1501, opened 2021-07-13
Sentry Integration

utku-ozturk:
Pls handle this PR witth https://github.com/4dn-dcic/shared-portal-components/pull/80

Trello: https://trello.com/c/vAop8tlJ

resim


src/encoded/static/components/app.js
```diff
@@ -223,6 +223,12 @@ export default class App extends React.PureComponent {
             );
         }
 
+        //Load up sentry io
```
alexkb0009, line 226: (See comments on SPC PR)

Thinking here or in SPC could essentially wrap in `if (typeof BUILDTYPE === 'string' && BUILDTYPE === "production") { ..initialize.. }` to avoid running on localhost/during-development (to reduce # of potential messages). (If/when not initialized, it could just pass thru to `console[whateverFunc](msg, args)`).

utku-ozturk: @alexkb0009 FF has a new configuration file, `sentry_config.json`, that contains dsn settings for the environments. Here, the localhost:wxyz and hotseat environments' values are empty. So logger will not log any messages into sentry unless those values are defined. Just to clarify, do we still need an extra check, such as `if (typeof BUILDTYPE === 'string' && BUILDTYPE === "production")` ?

``` json
{
    "hostnameDNSMapping" : {
        "data.4dnucleome.org" : "https:/[dns_url_for_data]",
        "mastertest.4dnucleome.org" : "https:/[dns_url_for_mastertest]",
        "staging.4dnucleome.org" : "https:/[dns_url_for_staging]",
        "fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com" : "",
        "hotseat.4dnucleome.org" : "",
        "localhost:8000" : "",
        "localhost:6543" : "",
        "default" : ""
    }
}
```

PR #1505, opened 2021-07-20
Static Content Added to Attribution Tab

utku-ozturk:
Trello: https://trello.com/c/76Kd4PNh

static content having "tab:attribution" are listed in attribution tab.

Screen Shot 2021-07-22 at 16 20 31

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "3.1.31"
+version = "3.1.32"
 description = "4DN-DCIC Fourfront"
 authors = ["4DN-DCIC Team <support@4dnucleome.org>"]
 license = "MIT"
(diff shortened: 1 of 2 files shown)


aschroed:
One question - is there a way to control where the StaticContent appears in relation to the stuff that is populated by default?

utku-ozturk:
> One question - is there a way to control where the StaticContent appears in relation to the stuff that is populated by default?

@aschroed Unfortunately, there is no out-of-box option to define static content's location with respect to Publications, Lab, and Submitter.

There are 2 possible options:

1. Current display order is `Static Content` - `Publications` - `Lab (+Award)` - `Submitted By`. We may move Static Content's location within the stack in the development phase. This option is not flexible and additionally, if there are multiple static sections, then all of them are rendered as a block. IMHO, it makes sense to display some of them on top, the rest of them on the bottom ..etc
2. Instead of single `tab:attribution` option, we may support `tab:attribution-top` or `tab:attribution-bottom` options. It requires some extra coding but provides more flexibility.

aschroed:
In this particular case we want the static content on the bottom - I think a reasonable and quick solution is to change the order of the stack to move the content to the bottom.  Looking at the existing case in for example the 'Publication' page overview tab extra static sections that are added to the Overview tab appear after the content that is automatically pulled in eg. https://data.4dnucleome.org/publications/0cbf34ae-fd3e-4b9d-a02f-cc80f1f8a872/ and there is not reason that the attribution tab should not behave in the same way.  In the future we may decide to add the flexibility you suggest but I don't think it is necessary to add that work right now.

PR #1506, opened 2021-07-21
Repair Auth0

willronchetti:
- Secret is no longer b64 encoded, so do not decode it


netsettler (approved):

netsettler:
Pondering whether this should be a major version bump, as the protocol for supplying the secret is incompatibly changed.

PR #1508, opened 2021-07-26
Upd upld key

aschroed:
Modifed the upload_key function to compare existing values stored in external propsheets and update if they are different than 
the calculated key.

Added a build_key class method that is used by this function as well as build_external_creds to make comparison possible.  

No explicit tests yet but the existing credential tests indicate nothing broken in build_external creds

### DIFF
--- a/src/encoded/tests/test_file.py
+++ b/src/encoded/tests/test_file.py
@@ -240,6 +240,16 @@ def test_files_aws_credentials_change_filename(testapp, fastq_uploading, file_fo
     assert res_put.json['@graph'][0]['href'].endswith('tiff')
 
 
+# def test_upload_key_modified_when_extcreds_no_match_key(testapp, fastq_uploading, file_formats):
+#     import pdb; pdb.set_trace()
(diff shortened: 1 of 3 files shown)


src/encoded/tests/test_file.py
```diff
@@ -135,12 +135,26 @@ def test_restricted_no_download(testapp, fastq_json):
     s3.put_object(Bucket='test-wfout-bucket', Key=resobj['upload_key'], Body=str.encode(''))
     download_link = resobj['href']
     testapp.get(download_link, status=307)
-    # fail download of restricted file (although with a 200 status?)
     testapp.patch_json(resobj['@id'], {'status': 'restricted'}, status=200)
+    # fail download of restricted file (although with a 200 status?)
```
willronchetti, line 139: This comment I think refers to the 200 status above (result of patching to restricted?). It behaves correctly in that it now 403's the download I'd say, so parens can probably be removed.

willronchetti (approved):

PR #1511, opened 2021-08-03
Micrometa App Container (Plain/Ajax)

utku-ozturk:
**Trello:** https://trello.com/c/6yahvAcl

Screen Shot 2021-08-03 at 20 49 06


alexkb0009 (approved):
Sorry I haven't had much time to catch up on reviews; code looks good and if working on dev environments then think is good to go.

PR #1512, opened 2021-08-05
Column's Initial Sort + Default Sort Fields

utku-ozturk:
Trello: https://trello.com/c/6iOGV8dz


alexkb0009:
Looks good to me would ask Will and maybe Sarah to check also since this mostly backend/schema stuff (if they approve then good to merge). Might be good to add a default/initial sort to one of the schemas as an example.

willronchetti (approved):
`search.py` changes look fine but needs version bump.

aschroed (approved):

andreacosolo (approved):

PR #1514, opened 2021-08-16
Bug Fix - MicroMeta App Bottom Bar Visibility

utku-ozturk:
Trello: https://trello.com/c/YnLrKruc

Micro Meta's bottom bar buttons are not visible in full-screen mode for some low-resolution displays.

Bug: https://i.gyazo.com/094a50766a1abf68a1719d828c4c0ad2.gif
Fix: https://i.gyazo.com/4dca5321115c473479178fbf8c8000c0.gif

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4291,9 +4291,9 @@
       },
       "dependencies": {
         "@babel/runtime-corejs2": {
-          "version": "7.14.6",
-          "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.14.6.tgz",
-          "integrity": "sha512-wICJNTtKkM0sB/Oip84Ul8W2lqMj/5eG5JsfgjfRnO69ucvSrwmVr2M2sS6ysV8s13EXwicEM0SuGRuFkvvbnQ==",
+          "version": "7.15.3",
+          "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.15.3.tgz",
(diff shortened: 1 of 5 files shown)


alexkb0009 (approved):
Looks decent .. I wonder if it might be possible/performant to pass in something along lines of `Math.min(windowHeight - (some hardcoded height of header), 300)` as a height prop to micrometa app to cause micrometa to resize itself in response to height of browser window. There's a lot of scrollbars visible in GIF which isn't super awesome in long term.. for the bottom one, maybe adding some padding like 20px to left and right of micrometa would help (some browsers/OSes render a scrollbar that takes up width, some don't; non-issue if app stretches to fit browser window always to point where no scrollbars, but if need scrollbars, then adding ~20px extra spacing at right (and maybe left, for symmetry) might help w. that).

utku-ozturk:
> Looks decent .. I wonder if it might be possible/performant to pass in something along lines of `Math.min(windowHeight - (some hardcoded height of header), 300)` as a height prop to micrometa app to cause micrometa to resize itself in response to height of browser window. There's a lot of scrollbars visible in GIF which isn't super awesome in long term.. for the bottom one, maybe adding some padding like 20px to left and right of micrometa would help (some browsers/OSes render a scrollbar that takes up width, some don't; non-issue if app stretches to fit browser window always to point where no scrollbars, but if need scrollbars, then adding ~20px extra spacing at right (and maybe left, for symmetry) might help w. that).

@alexkb0009 Yeah, you are definitely right. I added some padding to remove the unnecessary horizontal scrollbar. Moreover, I talked to AlexR about removing the hardware toolbar's horizontal scrollbar. I patched a hot-fix and release a new version including some typo updates from AlexR.

the latest look and feel is: https://i.gyazo.com/359113d0054d91c530e028cab305eb2f.gif

I'll merge it shortly.

PR #1515, opened 2021-08-17
Publication Title Multisort Option 

utku-ozturk:
Trello: https://trello.com/c/VxXCVNLN

A minor update to let users sorting Title column by Publication Title or Author Name.

PS: IMHO, the `authors` array field is not sorted by ES properly (https://data.4dnucleome.org/search/?type=Publication&sort=authors). Instead, should we update `search.py` to sort array fields by their first element?

Screen Shot 2021-08-17 at 20 05 44


alexkb0009 (commented):
Would move this into Publication schema probably under "columns"->"display_title" ([this is where we keep all the sort_fields in CGAP](https://github.com/dbmi-bgm/cgap-portal/search?q=sort_fields)). A good rule of thumb I think is if a column configuration/option is more 'data'-related and should be more easily editable by data wranglers then probably better to have it be in schema, but if is more presentational and should be primarily edited in context of the UI or by UI people (e.g. a custom render function, maybe width of column) then probably should be colExtensionMap.

utku-ozturk:
@alexkb0009 I moved `sort_fields` from PublicationView into `schema`.

alexkb0009 (approved):

aschroed (commented):
OK - there are a couple of things going on here.
First authors of a publication are specified in a particular order based on importance (i.e. who contributed most to the paper (earlier in the list) and who are the PIs/senior investigators in whose lab(s) the work was done (listed late/last in the list of authors).  Therefore, it is generally not meaningful and in fact removes information to sort the list of authors of a publication - i.e. the order should be maintained. 

What can be usefully sorted is the name of the first author listed for a publication - however, unfortunately due to oddities in how the names of authors formatted and automatically pulled into the data portal from external resources (PUBMED or BioRxiv) sometimes the surname is listed first (pubmed) and others the first name (Biorxiv) so the sort by first author won't be always be alphabetical by surname (which would be the desired behavior).  However, sort by first author name string can still be useful if not perfect and the sort by display_title should accomplish this - can you provide calcprops as sort_options?

utku-ozturk:
> Therefore, it is generally not meaningful and in fact removes information to sort the list of authors of a publication - i.e. the order should be maintained.

Sure, sorting the items within the authors' array is definitely not meaningful. I never considered it as an option.

> However, sort by first author name string can still be useful if not perfect and the sort by display_title should accomplish this - can you provide calcprops as sort_options?

Great! `display_title` takes 1st author into account. I'll test it shortly.
https://github.com/4dn-dcic/fourfront/blob/b383d89d543bc6db0536391c2dd7100ad9b22dc2/src/encoded/types/publication.py#L266-L292

aschroed:
Sounds good.  The first part of my comment was addressing your comment about the author sort not working correctly - which I agree appears to be a problem.  It looks like that Alex's suspicion that for this search - the authors values in the array are indeed sorted and then the results are returned based on the desc sort of those sorted items - which is why you see the first 4 results with 'Abdennur N' as an author regardless of the position of his name in the author list.  This requires a backend ticket but I wasn't sure if the idea of upgrading to ES7.0 would resolve this or if they were separate issues/tasks?

utku-ozturk:
@aschroed I deployed the recent changes into hotseat: http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?type=Publication IMO, sorting by authors works good but could you also check it when you are available?

utku-ozturk:
> Sounds good. The first part of my comment was addressing your comment about the author sort not working correctly - which I agree appears to be a problem. It looks like that Alex's suspicion that for this search - the authors values in the array are indeed sorted and then the results are returned based on the desc sort of those sorted items - which is why you see the first 4 results with 'Abdennur N' as an author regardless of the position of his name in the author list. This requires a backend ticket but I wasn't sure if the idea of upgrading to ES7.0 would resolve this or if they were separate issues/tasks?

Oh, I see. Yes, you are right. 'Abdennur N' example makes the ES's internal sort process clear. The custom array sorting seems like a back-end process but let me read some more articles on whether the sort script/workaround is available on ES6.

aschroed:
I agree that sorting by author now works as expected.  Thanks!

PR #1516, opened 2021-08-18
Bump some dependency versions

netsettler:
It was taking forever to `poetry update`, suggesting many libraries were out of date. 

I did some bumping to get it to where there were newer versions. I had to pin `botocore`, though, which gets tangled up with some `awscli` dependencies that conflict, according to `pip` and `poetry`. That, in turn, is caused by requiring `Submit4DN`.

Some questions that occur:

* Is `Submit4DN` in fact needed? Or could it just be loaded by someone planning to submit?
* If it is needed, could `awscli` be loaded on demand? Is it a requirement on the server?

Meanwhile, hopefully we can bump these other versions if tests passed. They passed on my machine other than a couple of  intermittents.

By the way, I also added a `clear-poetry-cache` target for `make` along the way.


willronchetti (approved):

willronchetti (approved):
Approving again with the test addition

PR #1524, opened 2021-09-16
Port NPM Dependecy Updates to 4DN

utku-ozturk:
**Trello:** https://trello.com/c/CT1Pm5OH


alexkb0009 (commented):
Looks good!
Would add dependency for "bootstrap": "^4.6.0" to package.json also. Right now 'bootstrap' happens to be a subdependency of 'react-bootstrap' so is available in `/node_modules/` but is not explicit or necessarily consistent (newer versions of React-Bootstrap might stop requiring it as dependency, etc.).

PR #1525, opened 2021-09-20
add protected_data to individual

andreacosolo:
Protected Data: when consent to publish sequencing data for an individual is not available, files are subject to restricted access.


src/encoded/schemas/individual_human.json
```diff
@@ -57,6 +57,13 @@
                 "postnatal"
             ]
         },
+        "protected_data": {
+            "title": "Protected Data",
+            "description": "A flag to indicate whether consent to publish sequence files from this individual is not available. If true, fastq and bam files are subject to restricted access.",
+            "type": "boolean",
+            "default": false,
```
aschroed, line 64: I wonder if it might make sense to have the default be true?  Safer that way and going forward it will require deliberate change for any wrangler creating a human individual?

aschroed (approved):
See my comment on the default for protected status and see what you think

andreacosolo:
[improve restricted files management](https://trello.com/c/xkCT0tM5/125-improve-restricted-files-management)

PR #1526, opened 2021-09-20
Repair Poetry

willronchetti:
- Bring poetry out of `1.1.4`, which is apparently broken now, to `1.1.9`
- Edit: the actual fix is to `setuptools`, which removed certain build functionality with version `58.0.0`


Makefile
```diff
@@ -43,7 +43,9 @@ macpoetry-install:  # Same as 'poetry install' except that on OSX Catalina, an e
 
 configure:  # does any pre-requisite installs
 	pip install --upgrade pip
-	pip install poetry==1.1.4  # this version is known to work, so we'll use it for now. -kmp 3-Mar-2021
+	pip install poetry==1.1.9  # this version is known to work, so we'll use it for now. -kmp 3-Mar-2021
```
netsettler, line 46: I would have just left it open, since 1.1.9 is latest and I'd _like_ to believe that it's now going to be maintained going forward, but I can live with this for now.

netsettler (approved):
My comment is not blocking.

PR #1530, opened 2021-09-23
Adjustments to allow acceptance of dcicutils 3.0.0

netsettler:
These are adjustments to be able to take dcicutils 3.0.0.
Presently everything is still in beta as I test things.

Included changes:

* Rewrite imported class `Deployer` as `BasicLegacyFourfrontIniFileManager`.
* Update some dependency requirements to match what `dcicutils` forces on us.
* Get rid of dependency on `s3transfer`, which should be implicit. We don't use this explicitly so should not pin it.


willronchetti (approved):
Should be deployed just in case.

netsettler:
Yeah, I had previously deployed this and it was working fine and testing fine. There seem to be some test glitches, but I don't think they are me. I think they are either transients or [recent changes that came in from master](https://github.com/4dn-dcic/fourfront/pull/1530/commits/52320211d2ab8eb6c30bc6540c231742fe4a3ba4).

PR #1533, opened 2021-09-28
Better test_index_data_workbook

netsettler:
* This starts by porting some test improvements and PEP8 from `cgap-portal`.
* That didn't entirely fix things, so I updated `test_index_data_workbook` to be more tolerant than what was ported in. I think this will make it less flaky. I'll port it back to `cgap-portal` sometime.
* There is a bunch of PEP8 in here.

I tried to add comments at each of the points of substance so they'd be easier to spot among the PEP8 changes.

### DIFF
--- a/src/encoded/tests/test_search.py
+++ b/src/encoded/tests/test_search.py
@@ -610,17 +610,6 @@ def test_index_data_workbook(app, workbook, testapp, indexer_testapp, htmltestap
     es = app.registry['elasticsearch']
     # we need to reindex the collections to make sure numbers are correct
     create_mapping.run(app, sync_index=True)
-    # re_index = False
-    # time.sleep(3)  # Give SQS and/or indexer a chance for things to subside
-    # indexer_status = testapp.get("/indexer_status?format=json").json
-    # for k, v in indexer_status.items():
-    #     if isinstance(v, int) and v > 0:
-    #         if k.startswith("dlq"):
-    #             raise AssertionError(f"DLQ is not empty: {indexer_status}")
-    #         else:
-    #             re_index = True
-    # if re_index:
-    #     testapp.post_json('/index', {})
     retried = False
     while True:
         # check counts and ensure they're equal
--- a/src/encoded/tests/test_search.py
+++ b/src/encoded/tests/test_search.py
@@ -1,6 +1,5 @@
 import json
 import pytest
-import time
 import webtest
 
 
@@ -22,11 +21,11 @@ pytestmark = [
     pytest.mark.schema,
     # pytest.mark.indexing,
     pytest.mark.workbook,
-    #pytest.mark.flaky(rerun_filter=customized_delay_rerun(sleep_seconds=10))
+    # pytest.mark.flaky(rerun_filter=customized_delay_rerun(sleep_seconds=10))
 ]
 
 
-### IMPORTANT
+# == IMPORTANT ==
 # uses the inserts in ./data/workbook_inserts
 # design your tests accordingly
 notice_pytest_fixtures(app_settings, app, workbook)
@@ -47,8 +46,9 @@ def recursively_find_uuids(json, uuids):
 
 
 def test_search_view(workbook, testapp):
+    notice_pytest_fixtures(workbook)
     res = testapp.get('/search/?type=Item').json
-    assert res['@type'] == ['ItemSearchResults','Search']
+    assert res['@type'] == ['ItemSearchResults', 'Search']
     assert res['@id'] == '/search/?type=Item'
     assert res['@context'] == '/terms/'
     assert res['notification'] == 'Success'
@@ -60,6 +60,7 @@ def test_search_view(workbook, testapp):
 
 
 def test_search_with_no_query(workbook, testapp):
+    notice_pytest_fixtures(workbook, testapp)
     # using /search/ (with no query) should default to /search/?type=Item
     # thus, should satisfy same assertions as test_search_view
     res = testapp.get('/search/').follow(status=200)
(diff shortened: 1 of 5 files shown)


src/encoded/tests/test_search.py
```diff
@@ -543,22 +555,83 @@ def test_collection_actions_filtered_by_permission(workbook, testapp, anontestap
     assert len(res.json['@graph']) == 0
 
 
-@Retry.retry_allowed('test_index_data_workbook.check', wait_seconds=1, retries_allowed=5)
-def check_item_type(client, item_type):
-    # This might get a 404 if not enough time has elapsed, so try a few times before giving up.
-    return client.get('/%s?limit=all' % item_type, status=[200, 301]).follow()
+class ItemTypeChecker:
```
netsettler, line 558: This class is imported from `cgap-portal`.

src/encoded/tests/test_search.py
```diff
@@ -543,22 +555,83 @@ def test_collection_actions_filtered_by_permission(workbook, testapp, anontestap
...
+        # We retry a lot of times because it's still fast if things are working quickly, but if it's
+        # slow it's better to wait than fail the test. Slowness is not what we're trying to check for here.
+        # And even if it's slow for one item, that same wait time will help others have time to catch up,
+        # so it shouldn't be slow for others. At least that's the theory. -kmp 27-Jan-2021
+        extra = "&status=deleted" if deleted else ""
+        return client.get('/%s?limit=all%s' % (item_type, extra), status=[200, 301]).follow()
+
+
+    CONSIDER_DELETED = True
+    DELETED_SEEN = {}
```
netsettler, line 574: DELETED_SEEN is a cache that keeps track for each item type of how many deleted have been seen.
There is no real need to reset the cache because we use it in situations where a test maps across all types and updates the values.

src/encoded/tests/test_search.py
```diff
@@ -543,22 +555,83 @@ def test_collection_actions_filtered_by_permission(workbook, testapp, anontestap
...
+                items_not_deleted = []
+            try:
+                res = cls.check_item_type(client, item_type, deleted=True)
+                items_deleted = res.json.get('@graph', [])
+            except webtest.AppError:
+                items_deleted = []
+            if items_deleted:
+                cls.DELETED_SEEN[item_type] = items_deleted
+            else:
+                cls.DELETED_SEEN.pop(item_type, None)  # delete entry if present but we want it empty
```
netsettler, line 604: This keeps DELETED_SEEN up-to-date.

src/encoded/tests/test_search.py
```diff
@@ -596,7 +671,49 @@ def test_index_data_workbook(app, workbook, testapp, indexer_testapp, htmltestap
                 html_res = htmltestapp.get(item_res['@id'])
                 assert html_res.body.startswith(b'<!DOCTYPE html>')
             except Exception as e:
+                ignored(e)
                 pass
+    if ItemTypeChecker.CONSIDER_DELETED:
+        print(f"(CONSIDER_DELETED) Items deleted = {ItemTypeChecker.DELETED_SEEN}")
+        print(f"db_total={db_total} es_total={es_total} deleted_seen={ItemTypeChecker.deleted_seen()}")
+        assert(db_total == es_total + ItemTypeChecker.deleted_seen())  # 2nd is db, 4th is es
```
netsettler, line 679: Here we accommodate deleted items. They will be present in situations where `ItemTypeChecker.DELETED_ITEMS` is `True`, so we use that as a conditional.  (Some tests bind this class variable using `with local_attrs(): ...`)

src/encoded/tests/test_search.py
```diff
@@ -596,7 +671,49 @@ def test_index_data_workbook(app, workbook, testapp, indexer_testapp, htmltestap
...
+    else:
+        print(f"(not CONSIDER_DELETED) db_total={db_total} es_total={es_total} deleted_seen={ItemTypeChecker.deleted_seen()}")
+        assert(db_total == es_total)  # 2nd is db, 4th is es
+
+@pytest.mark.manual
+@pytest.mark.skip
+def test_index_data_workbook_after_posting_deleted_page_c4_570(workbook, testapp, html_testapp):
```
netsettler, line 686: This test has to be invoked manually. It takes a very long time to run, but is potentially useful for debugging. It was imported from `cgap-portal`, where the bug it is a regression test for was found.

src/encoded/tests/test_search.py
```diff
@@ -543,22 +555,83 @@ def test_collection_actions_filtered_by_permission(workbook, testapp, anontestap
...
+    retried = False
+    while True:
+        # check counts and ensure they're equal
+        testapp_counts = testapp.get('/counts')
+        # e.g., {"db_es_total": "DB: 748 ES: 748 ", ...}
+        db_es_total = testapp_counts.json['db_es_total']
+        split_counts = db_es_total.split()
+        db_total = int(split_counts[1])
+        es_total = int(split_counts[3])
+        if db_total == es_total or retried:
```
netsettler, line 626: We used to just do an assert of this, but that made the test end before we got better data. It also meant that if the counts just hadn't stabilized, we'd lose falsely, so now we go ahead and loop once. That seems to be what fixes this to be a more stable test.

src/encoded/tests/workbook_fixtures.py
```diff
@@ -67,6 +72,8 @@ def workbook(app):
     if load_res:
         raise(load_res)
 
+    time.sleep(3)  # Fixture is time-consuming anyway, so indulge a sleep to let SQS internals to catch up
```
netsettler, line 75: This is a small but non-cosmetic change. Since this fixture is created only once, it shouldn't add much to overall testing.

willronchetti (approved):

netsettler:
This PR tests cleanly on a local machine, so I think the GA issue is not related to the test.
Also this doesn't change anything but tests, so I feel confident it's safe even though the GA testing has issues.
Going to go ahead and merge.

PR #1540, opened 2021-10-08
Containerize

willronchetti:
- Adds support for a containerized deployment of fourfront
- The vast majority of this update is identical to what already exists in CGAP
- Mastertest: http://fourfront-mastertest-1312345630.us-east-1.elb.amazonaws.com/health
- Green: http://fourfront-productiongreen-1285340631.us-east-1.elb.amazonaws.com/health
- Blue: http://fourfront-productionblue-1843150337.us-east-1.elb.amazonaws.com/health


deploy/docker/production/entrypoint.bash
```diff
@@ -0,0 +1,38 @@
...
+# the "$application_type" environment variable. Possible options are:
+#  * "deployment" to run the deployment
+#  * "indexer" to run the production indexer (forever)
+#  * "portal" to run the production portal worker (API back-end)
+#  * "local" to run a local deployment
+
+# Note that only "local" can be run from the local machine
+# but the same image build is run across the entire local/production stack.
+
+deployment="deployment"
```
netsettler, line 15: What's the advantage of doing this over just referencing "deployment" below? In Python it has virtue because you'd get an unbound variable error if you misspelled something. In a shell script you'll just end up comparing to the empty string if you make a typo. I guess it _vaguely_ provides some doc of what states are allowed, but you could just do that in a comment or tell people to read the if statement below.

deploy/docker/postgres/postgresql.conf
```diff
@@ -0,0 +1,642 @@
...
+
+
+#------------------------------------------------------------------------------
+# CONNECTIONS AND AUTHENTICATION
+#------------------------------------------------------------------------------
+
+# - Connection Settings -
+
+# Only necessary change from defaults
+listen_addresses = '*'
```
netsettler, line 60: Why `'*'`? [The doc for `listen_addresses`](https://postgresqlco.nf/doc/en/param/listen_addresses/) says:

> While client authentication (client-authentication) allows fine-grained control over who can access the server, listen_addresses controls which interfaces accept connection attempts, which can help prevent repeated malicious connection requests on insecure network interfaces. [...]
> 
> RECOMMENDATIONS
> Set your listen_address as restrictively as possible; '*' should only be used for development machines

package.json
```diff
@@ -109,6 +109,7 @@
     "d3": "^5.16.0",
     "detect-browser": "^3.0.1",
     "domready": "^0.3.0",
+    "fbjs": "^0.8.17",
```
alexkb0009, line 112: What's this be for?

alexkb0009: Oh yeah I think we discussed; this pinned fbjs was able to be removed on CGAP after [updating html-react-parser dependencies](https://github.com/dbmi-bgm/cgap-portal/pull/475). Not sure if same still needs to be done on 4DN-side... probably updated version is already on here since most of changes we in shared-portal-components.

The things that might need to do on 4DN-side is [update html-react-parser version](https://github.com/dbmi-bgm/cgap-portal/pull/475/commits/91a2d6ef2f8eadb8f49f0677ac709e63e7313229) (4DN is still on [old version in its package.json](https://github.com/4dn-dcic/fourfront/blob/master/package.json#L116)) and ~~[these minor updates to webpack.config.js](https://github.com/dbmi-bgm/cgap-portal/pull/475/commits/dd2e017e7d1c99285e26c9ffdfad070c82b7d1e0#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166f)~~ (nvm -- these look like are in master already). (@utku-ozturk)

deploy/docker/production/entrypoint_deployment.bash
```diff
@@ -9,11 +9,11 @@ poetry run python -m assume_identity
 # Clear db/es on fourfront-mastertest if we run an "initial" deploy
 # Do nothing on other environments
 if [ -n "${INITIAL_DEPLOYMENT}" ]; then
-  poetry run clear-db-es-contents production.ini --app-name app --env fourfront-mastertest
+  poetry run clear-db-es-contents production.ini --app-name app --env fourfront_mastertest
```
netsettler, line 12: Should this also work by environment variable with the env name in a variable? Is it already in `ENV_NAME`?

deploy/docker/production/nginx.conf
```diff
@@ -80,13 +100,33 @@ http {
...
             proxy_set_header  X-Forwarded-For    $proxy_add_x_forwarded_for;
-            proxy_set_header  X-Forwarded-Proto  $scheme;
-            proxy_pass  http://app;
             proxy_set_header  Connection  "";
+
+            # "ifs are evil" except for when you need them
+            # the below code is a "hack" for doing "if cond1 AND cond2" - Will Oct 14 21
+            # if we are a cgap*.hms.harvard.edu host
+            set $is_https_host "";
+            if ($host ~* (cgap.*\.hms\.harvard\.edu)) {
```
netsettler, line 113: Do you now want to add a hyphen after cgap to disallow the original `cgap.hms.harvard.edu`? Are we going to be using that name again? (Maybe if not we should leave it up with forwarding info.) Or is it only `fourfront-cgap` we are discontinuing?

src/encoded/commands/load_access_keys.py
```diff
@@ -54,7 +56,12 @@ def generate_access_key(testapp, env, user_uuid, description):
     Returns:
         dict: access key contents with server
     """
-    server = get_beanstalk_real_url(env)
+    try:
```
netsettler, line 68: Do we really not have any manifest in-memory notion of whether we expect an ECS or EB url so that we can just make this an ordinary `if` and not a `try`?

src/encoded/commands/load_access_keys.py
```diff
@@ -83,6 +90,8 @@ def main():
     )
     parser.add_argument('config_uri', help='path to configfile')
     parser.add_argument('--app-name', help='Pyramid app name in configfile')
+    parser.add_argument('--secret-name', help='name of application identity stored in secrets manager within which'
```
netsettler, line 103: Why `--secret-name` and not `--s3-encrypt-key`?

netsettler (approved):
My comments are not blocking.

PR #1549, opened 2021-11-03
use doi.org and biorxiv.org APIs

andreacosolo:
Biorxiv seems to require using its API for programmatic access. Requests to doi.org were failing because it was redirecting to biorxiv.

Instead, use the doi.org API (no redirects) to determine whether the doi points to biorxiv, and then use the biorxiv API to get the Publication metadata.

### DIFF
--- a/src/encoded/types/publication.py
+++ b/src/encoded/types/publication.py
@@ -132,7 +132,7 @@ def fetch_biorxiv(url, doi):
     record_dict = r.json()['collection'][0]
     pub_data = {k: record_dict.get(v) for k,v in fdn2biorxiv.items() if record_dict.get(v)}
     if pub_data.get('authors'):  # format authors according to 4DN schema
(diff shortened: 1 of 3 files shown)


src/encoded/types/publication.py
```diff
@@ -175,14 +159,18 @@ def map_doi_pmid(doi):
...
         if resp.status_code == 200:
             break
-    landing_page = resp.url
-    if "biorxiv" in landing_page.lower():
+    if resp.json().get('values'):
+        for value in resp.json()['values']:
+            if value.get('type', '') == 'URL':
+                landing_page = value['data']['value']
+                break
+    if "biorxiv" in landing_page:
```
aschroed, line 173: No need to lowercase as before?

aschroed (approved):

src/encoded/types/publication.py
andreacosolo: No, this is now the URL to which doi.org redirects, and 'biorxiv' here is already lowercase.

PR #1560, opened 2021-11-12
MicroMeta Hardware Summary Table

utku-ozturk:
Pls handle this PR with its [SPC pair](https://github.com/4dn-dcic/shared-portal-components/pull/90).

Trello: https://trello.com/c/usRCpJld

Screen Shot 2021-11-23 at 22 36 43

Screen Shot 2021-11-23 at 22 39 29

Screen Shot 2021-11-23 at 22 38 26


src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
```diff
@@ -574,6 +591,457 @@ export class MicroMetaTabView extends React.PureComponent {
...
+        return {
+            'tab': <span><i className="icon icon-list-alt fas icon-fw" /> Hardware Summary</span>,
+            'key': 'hardware-summary',
+            'content': <MicroMetaSummaryTabViewFRef {...props} width={width} />
+        };
+    }
+
+    static facetsFromMicroscope(microscope) {
+        const categoryObj = {};
+        _.forEach(microscope.components, function (component) {
```
alexkb0009, line 610: (For future) Using native version of e.g. Array.forEach (+ map, Object.keys, etc.) is slightly preferred, in case we want to migrate away from using underscore at some point. Previously, some browsers didn't implement their own .forEach, .map, .reduce, etc. in their JS engines/implementations, but this was 5-10+ yrs ago. There's some other handy underscore functions which we use still though (zip, throttle, debounce, pluck, ...), though theoretically we could replace with lighter-weight subset of lodash or something at some point in future.

alexkb0009:
Code looks great! Will test a tiny bit shortly and then approve.

alexkb0009:
Is this on hotseat? Browsed around a bit on e.g. [here](http://fourfront-hotseat.9wzadzju3p.us-east-1.elasticbeanstalk.com/microscope-configurations/d63949c2-e357-4b77-9b89-8f8de1c025de/#micrometa) but couldn't get to the view of component listings?

Am booting up local env right now but unsure if I have any test inserts.

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
```diff
@@ -574,6 +591,457 @@ export class MicroMetaTabView extends React.PureComponent {
...
+            tableBody = _.map(_.keys(schema.subCategoriesOrder), function(subCategory){
+                const subCategoryProperties = _.filter(schemaPropPairs, function (spp) {
+                    const [, propItem] = spp;
+                    return propItem && propItem.category === subCategory;
+                });
+                const sectionProps = {
+                    subCategory, subCategoryProperties, matches: visibleMatches, tooltipLimit, columClassName, 
+                    collapsed: !!collapsedSections[subCategory]
+                };
+                return <CollapsibleSubCategory key={subCategory} {...sectionProps} toggleExpand={this.toggleExpand} />
```
alexkb0009, line 942: ~~`this.toggleExpand` here refers to `undefined` because not using an arrow function (which would inherit `this`), a raw function 's `this` refers to itself~~ ah I see is using the _.map 2nd arg. Would still use an arrow function for clarity, though, and because might migrate away from underscore in future.

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
```diff
@@ -574,6 +591,457 @@ export class MicroMetaTabView extends React.PureComponent {
...
+            title: "Hardware Explorer",
+            facets,
+            onFilter: this.onFilter,
+            useRadioIcon: true,
+            persistSelectedTerms: false
+        };
+
+        return (
+            <div class="overflow-hidden">
+                <h3 class="tab-section-title"><span>Hardware Summary</span></h3>
```
alexkb0009, line 974: Need to use `className`, not `class`.

(Do you guys have ESLint or any other linting setup on VSCode? It can help catch small stuff like this).

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
```diff
@@ -33,22 +38,32 @@ export default class MicroscopeConfigurationView extends DefaultItemView {
...
         return initTabs.concat(this.getCommonTabs()); // Add remainder of common tabs (Details, Attribution)
     }
 
 }
 
+const MicroMetaTabViewFRef = React.forwardRef((props, ref) => <MicroMetaTabView {...props} forwardRef={ref} />);
 export class MicroMetaTabView extends React.PureComponent {
 
     static getTabObject(props, width) {
+        const { ref } = props;
```
alexkb0009, line 61: Is unused

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
```diff
@@ -574,6 +591,457 @@ export class MicroMetaTabView extends React.PureComponent {
...
+
+        return facets;
+    }
+
+    static defaultLayoutSettings(windowWidth, mounted, matches) {
+        let length = 1;
+        if (matches) {
+            if (Array.isArray(matches)) {
+                length = matches.length;
+            } else if (typeof matches === 'int' || typeof matches === 'number') {
```
alexkb0009, line 646: 'int' not a valid JS type, use 'number'.

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
alexkb0009: (Updated/pushed)

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
alexkb0009: (Updated to 'className' / pushed)

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
alexkb0009: (Updated/pushed)

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
alexkb0009: (Updated/pushed)

alexkb0009:
(Low priority / not too important)

A [line appears at bottom of Facet](https://gyazo.com/96e5045c7ebcaae3ddc923b859618fc8) when select the last term only, on some Facets. Doesn't hold up the PR but something to look into when have a chance.

alexkb0009 (approved):

src/encoded/static/components/item-pages/MicroscopeConfigurationView.js
utku-ozturk: @alexkb0009 My bad, it looks that my ESLint is failing for a while since I moved my working folder and VS Code is not matching with folder name in .vscodee/setting.json. Thank you.

utku-ozturk:
> (Low priority / not too important)
> 
> A [line appears at bottom of Facet](https://gyazo.com/96e5045c7ebcaae3ddc923b859618fc8) when select the last term only, on some Facets. Doesn't hold up the PR but something to look into when have a chance.

Good point! we have fixed it in SPC release 0.1.29.

PR #1568, opened 2021-12-06
Link to Microscope Configuration in Experiment/File View

utku-ozturk:
Trello: https://trello.com/c/NKCfQAOY

1. Microscope Configuration is added into the `Properties` section on `Experiment` page.
2. Microscope Configuration column is added into Raw Files table in Experiment/ExperimentSet pages. The column displays a microscope icon and the column is hidden unless at least one raw file is associated with a configuration.
3. Microscope Configuration is added into the `Overview` tab on the `Microscopy File` page.
4. New memoized `allRawFiles` function is added to increase performance since Notes, QC, Microscope Configuration columns check the existence of relevant fields in raw files.
5. `showNotesColumns` prop was used to be a string ("never", "conditional", "always") but it is not so performant and prone to mistyping errors. So it is converted to bool (true, false, null/undefined for conditional case)
6. Additionally, missing `files.notes_to_tsv` and `files.file_type` embeds added. (extra)

Screen Shot 2021-12-05 at 23 55 15

Screen Shot 2021-12-06 at 17 16 47

### DIFF
--- a/deploy/post_deploy_testing/cypress/integration/01c_static_content_joint_analysis.js
+++ b/deploy/post_deploy_testing/cypress/integration/01c_static_content_joint_analysis.js
@@ -111,7 +111,7 @@ describe('Joint Analysis Page', function () {
                     expect(nextTotalCount).to.be.greaterThan(origTotalCount);
                     expect(nextTotalCount).to.be.greaterThan(49);
                 });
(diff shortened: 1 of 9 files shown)


alexkb0009 (approved):
Looks great!

PR #1572, opened 2022-01-10
Top Search Bar Redesign

utku-ozturk:
**Trello:** https://trello.com/c/lBg8kPJ8

1. Top Search Bar's single search icon is replaced with search text box to make it more prominent.

resim

2. When a user clicks the search text in the top search bar, a form that includes a search type dropdown and a text box is displayed in the big drop-down menu. 

resim

3. While Experiment Sets navigates search to /browse, the other search types' endpoint is a /search query. (By Accession is an exception, that navigates to the Item page directly.)

Screen Shot 2022-01-10 at 16 54 26

4. Mobile display
Screen Shot 2022-01-10 at 16 56 32

### DIFF
--- a/src/encoded/static/components/navigation/components/LeftNav.js
+++ b/src/encoded/static/components/navigation/components/LeftNav.js
@@ -1,8 +1,11 @@
 'use strict';
 
-import React, { useMemo } from 'react';
+import React, { useMemo, useState, useCallback } from 'react';
 import Nav from 'react-bootstrap/esm/Nav';
+import DropdownItem from 'react-bootstrap/esm/DropdownItem';
+import DropdownButton from 'react-bootstrap/esm/DropdownButton';
 import url from 'url';
+import _ from 'underscore';
 import { console, memoizedUrlParse } from '@hms-dbmi-bgm/shared-portal-components/es/components/util';
 import { navigate } from './../../util'; // Extended w. browseBaseHref & related fxns.
 import {
@@ -23,6 +26,7 @@ export const LeftNav = React.memo(function LeftNav(props){
             <ToolsNavItem {...props} />
             <ResourcesNavItem {...props} />
             <HelpNavItem {...props} />
+            <SearchNavItem {...props} />
         </Nav>
     );
 });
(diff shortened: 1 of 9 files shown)


src/encoded/static/components/navigation/components/LeftNav.js
```diff
@@ -166,3 +171,181 @@ const DataNavItemBody = React.memo(function DataNavItemBody(props) {
...
+        }
+    });
+    //navigate to Item page directly without searching
+    const navigateByAccession = useCallback(function (evt) {
+        if (searchItemType === 'ByAccession') {
+            evt.preventDefault();
+            evt.stopPropagation();
+            navigate('/' + searchText);
+        }
+    });
```
alexkb0009, line 271: `useCallback(function eventHandlerFunction(e){ ... } })` is a shorthand form of `useMemo(function(){ return eventHandlerFunction(e){ ... }; })`. So probably need to have `[ searchItemType ]` as 2nd param for these, otherwise `getSearchTextPlaceholder` might always return same thing even if value of searchItemType has changed. To improve performance could theoretically use a useMemo and return multiple functions from it so that 'searchItemType' doesn't have to get compared for every function, i.e. 

```
const { getSearchTextPlaceholder, navigateByAccession } = useMemo(function(){
    return {
        getSearchTextPlaceholder: function () {
            switch (searchItemType) {
                case 'Item':
                    return 'Search 4DN Data Portal';
                case 'ByAccession':
                    return 'Type Item\'s Accession (e.g. 4DNXXXX ...)';
                default:
                    return "Search in " + AvailableSearchItemTypes[searchItemType].text;
            }
        },
        navigateByAccession: function (evt) {
            if (searchItemType === 'ByAccession') {
                evt.preventDefault();
                evt.stopPropagation();
                navigate('/' + searchText);
            }
        },
        ........
    };
}, [ searchItemType ]);
```

src/encoded/static/components/navigation/components/LeftNav.js
```diff
@@ -166,3 +171,181 @@ const DataNavItemBody = React.memo(function DataNavItemBody(props) {
...
+    //select all text when focused
+    const handleFocus = useCallback(function (evt) {
+        evt.target.select();
+    });
+
+    const selectedItem = AvailableSearchItemTypes[searchItemType];
+    const action = (selectedItem && selectedItem.action) || '/search';
+    const btnIconClassName = 'icon icon-fw fas ' + (searchItemType === 'ByAccession' ? 'icon-arrow-right' : 'icon-search');
+    const btnDisabled = !(searchText &&  typeof searchText === 'string' && searchText.length > 0);
+    const searchTextClassName = 'form-control search-query w-100' + (!searchInputIsValid ? ' border border-danger' : '');
```
alexkb0009, line 294: Instead of 'w-100', would get rid of ".form-inline" className from the ancestor `<form>` element. If checkout the CSS definitions, '.form-control' by default is display: block, however form-inline turns form-controls into inline inputs instead of block ones, and thus have to use w-100 or similar to make width stretch, whereas display: block is 100% width by default and would get rid of this issue as well as code/complexity.

src/encoded/static/components/navigation/components/LeftNav.js
```diff
@@ -166,3 +171,181 @@ const DataNavItemBody = React.memo(function DataNavItemBody(props) {
...
+
+    return (//Form submission gets serialized and AJAXed via onSubmit handlers in App.js
+        <React.Fragment>
+            <h4>Search</h4>
+            <form action={action} method="GET" className="form-inline navbar-search-form-container" onSubmit={navigateByAccession}>
```
alexkb0009, line 299: I.e. can get rid of form-inline here. Form-inline is for if you are going to lay out the children as display: inline items, but you're using flex (via className="row"), so inline is unnecessary.

src/encoded/static/components/navigation/components/LeftNav.js
```diff
@@ -12,6 +15,7 @@ import {
     BigDropdownPageTreeMenuIntroduction,
     BigDropdownBigLink
 } from './BigDropdown';
+import { SearchBar } from '.';
```
alexkb0009, line 18: Would write as `import { SearchBar } from './SearchBar';`
If are going to ".", then is going to "./index.js" to grab SearchBar from there, and such patterns can cause glitches in rare cases (i.e. LeftNav is trying to load file that exports LeftNav also; if LeftNav is imported somewhere before SearchBar, then it might throw an Exception).

Is also more performant to import direct files (rather than from index.js) where possible for performance, because it helps inform [WebPack Tree-Shaking](https://webpack.js.org/guides/tree-shaking/) to skip files that aren't imported at all. For example we do `import DropdownButton from 'react-bootstrap/esm/DropdownButton';` instead of `import DropdownButton from 'react-bootstrap';` now because it allows webpack to get rid of all of the React-Bootstrap components that we don't actually use from the compiled bundle.js/server.js.

src/encoded/static/components/navigation/components/LeftNav.js
```diff
@@ -166,3 +171,181 @@ const DataNavItemBody = React.memo(function DataNavItemBody(props) {
...
+
+    const selectedItem = AvailableSearchItemTypes[searchItemType];
+    const action = (selectedItem && selectedItem.action) || '/search';
+    const btnIconClassName = 'icon icon-fw fas ' + (searchItemType === 'ByAccession' ? 'icon-arrow-right' : 'icon-search');
+    const btnDisabled = !(searchText &&  typeof searchText === 'string' && searchText.length > 0);
+    const searchTextClassName = 'form-control search-query w-100' + (!searchInputIsValid ? ' border border-danger' : '');
+
+    return (//Form submission gets serialized and AJAXed via onSubmit handlers in App.js
+        <React.Fragment>
+            <h4>Search</h4>
```
alexkb0009, line 298: Would look into reusing BigDropDownHeader or whatever title component might exist already.
Biggest thing from UI/styling perspective to improve would be to make the ['Search' title](https://gyazo.com/6777153998571ed052926b630c484d8b) match the [existing titles](https://gyazo.com/163f3e91163684cb3294334adbcd0a04) more closely, maybe have a large magnifying glass icon next to it or something (or not, depends on how it looks).

PR #1588, opened 2022-01-31
Address problems with deprecated AcceptValidHeader.best_match (C4-627)

netsettler:
This PR ports various bits of support from cgap-portal in order to address a problem with continued use of deprecated method `AcceptValidHeader.best_match` (See [C4-627](https://hms-dbmi.atlassian.net/browse/C4-627) for details.)

Note that there are unit tests in place that were already passing before this change was made and that check 18114 cases of different parameter combinations to make sure the behavior of these heavily used functions stay stable. I had to make some adjustments because the `cgap-portal` code exactly as expressed broke 384 of these cases. But because the unit test had good error diagnostics, it was clear what part was causing the problem, and for now I have favored bug-for-bug compatibility and left the behavior in fourfront "as-is" so that the old test now passes correctly. This narrows the differences between `cgap-portal` and `fourfront` in the `renderers.py` file, and enables a useful discussion about which way we should tip things to bring these two files into better alignment. It may in fact be that `cgap-portal` needs to change, but we can talk about that another day.

For now, the good outcome of this change is that it phases out uses of `.best_match` (as long as the `MIME_TYPE_TRIAGE_MODE` variable is left with a value of `modern`), and that means we won't get a zillion warnings from the unit tests, since this deprecated would be called at least once per unit test for almost all unit tests, resulting in quite a lot of warnings. Not to mention this method really _will_ go away at some point, and it's good to be rid of it.

Fixing this will also help the Python 3.7 effort, though it's good to do even for Python 3.6 because those warnings (described in [C4-627](https://hms-dbmi.atlassian.net/browse/C4-627)) _are_ annoying.

Note, too, that there were some other attempts at a PR to address this that were broader in scope and that didn't get patched for some reason, and grew stale. ([PR #1455](https://github.com/4dn-dcic/fourfront/pull/1455) and [PR #1465](https://github.com/4dn-dcic/fourfront/pull/1465)). I think those can be ignored for now, though I may go back and try to close them out or make them less stale if this PR, which is narrower, is approved.


src/encoded/tests/test_renderers.py
```diff
@@ -20,8 +25,53 @@ def __init__(self, content_type=None, status_code: int = 200, json=None, content
         super().__init__(status_code=status_code, json=json, content=content, url=url)
 
 
+def test_mime_variables():
```
netsettler, line 28: Note that this and a couple other unit tests are added in this file, but importantly, the function `test_should_transform` (on [original lines 39-136](https://github.com/4dn-dcic/fourfront/pull/1588/files#diff-a8b97191f716cd1392c16aa1d416fbd43a15116d8d4ce6764a1dbfcfb139f892L39-L136)) are unchanged. This gives pretty good confidence that the core functionality being tested here is unaffected in spite of the reimplementation of some functions used therein.

src/encoded/renderers.py
```diff
@@ -196,20 +209,6 @@ def security_tween(request):
 
         return handler(request)
 
-        # This was commented out when we introduced JWT authentication
```
netsettler, line 199: This is just getting rid of code that's long been commented out and is not believed to be needed now.

src/encoded/renderers.py
```diff
@@ -314,14 +316,81 @@ def canonical_redirect(event):
...
 
+# Web browsers send an Accept request header for initial (e.g. non-AJAX) page requests
+# which should contain 'text/html'
+MIME_TYPE_HTML = 'text/html'
+MIME_TYPE_JSON = 'application/json'
+MIME_TYPE_LD_JSON = 'application/ld+json'
+
+# Note: In cgap-portal, MIME_TYPE_JSON is at the head of this list. In fourfront, MIME_TYPE_HTML is.
+# The cgap-portal behavior might be a bug we should look at bringing into alignment. -kmp 29-Jan-2022
+MIME_TYPES_SUPPORTED = [MIME_TYPE_HTML, MIME_TYPE_JSON, MIME_TYPE_LD_JSON]
```
netsettler, line 327: This line is intentionally materially different between `cgap-portal` and `fourfront`. When I rewrote this code for `cgap-portal`, as indicated in the comment, I erroneously made the first element of this list be `MIME_TYPE_JSON`. This wasn't quite right, though is mostly harmless, but it represented a slight incompatible change for `cgap-portal` that I did not intend. I think the right thing may be to fix `cgap-portal` in the future, but for now this just highlights the essential difference between the two implementations. A cascade effect of this is that `MIME_TYPE_DEFAULT` on the next line ends up being `MIME_TYPE_HTML` here but `MIME_TYPE_JSON` in `cgap-portal`. That's probably wrong as well.

This keeps `test_should_transform` passing unmodified. However, some accommodations to newly ported unit tests in `test_renderers.py` had to be made to accommodate this shift. I think they're pretty unobjectionable changes, though.

src/encoded/renderers.py
```diff
@@ -331,6 +400,22 @@ def should_transform(request, response):
     if response.content_type != 'application/json':
         return False
 
+    # cgap uses the following extra check that I think is the right thing, but leaving this commented out
```
netsettler, line 403: This is the other significant difference between `cgap-portal` and `fourfront`. See the comment text for details. This makes `test_should_transform` in `test_renderers.py` pass properly.

src/encoded/tests/test_renderers.py
```diff
@@ -20,8 +25,53 @@ def __init__(self, content_type=None, status_code: int = 200, json=None, content
...
+    assert set(MIME_TYPES_SUPPORTED) == {MIME_TYPE_JSON, MIME_TYPE_HTML, MIME_TYPE_LD_JSON}
+    # Check that the first element is consistent with the MIME_TYPE_DEFAULT.
+    # It's an accident of history that this next relationship matters, but at this point check for consistency.
+    assert MIME_TYPE_DEFAULT == MIME_TYPES_SUPPORTED[0]
+    # Now we concern ourselves with the actual values...
+    # TODO: I think it's a bug that JSON is at the head of this list (and so the default) in cgap-portal.
+    #       cgap-portal needs to be made to match what Fourfront does to dig it out of a bug I introduced.
+    #       -kmp 29-Jan-2022
+    assert MIME_TYPES_SUPPORTED == [MIME_TYPE_HTML, MIME_TYPE_JSON, MIME_TYPE_LD_JSON]
+    assert MIME_TYPE_DEFAULT == MIME_TYPE_HTML
```
netsettler, line 47: These two lines in this test are different than in `cgap-portal` because they have to correspond to differences I had to make in `renderers.py` to make this a non-breaking change.

src/encoded/tests/test_util.py
```diff
@@ -0,0 +1,97 @@
...
+    # compute_set_difference_one, find_other_in_pair,
+    delay_rerun,  # utc_today_str,
+    customized_delay_rerun, check_user_is_logged_in
+)
+
+
+pytestmark = pytest.mark.working
+
+
+# def test_compute_set_difference_one():
```
netsettler, line 16: This is just opportunistic. These functions are not used any more.

src/encoded/tests/test_util.py
```diff
@@ -0,0 +1,97 @@
...
+    custom_delay = 0.5
+    half_delay_rerun = customized_delay_rerun(sleep_seconds=custom_delay)
+    t0 = datetime.datetime.now()
+    half_delay_rerun()
+    t1 = datetime.datetime.now()
+    assert (t1 - t0).total_seconds() > custom_delay
+    assert (t1 - t0).total_seconds() < custom_delay + DELAY_FUZZ_SECONDS
+
+
+# def test_utc_today_str():
```
netsettler, line 75: We get this function from `dcicutils` these days so don't need it here.

src/encoded/util.py
```diff
@@ -296,31 +296,31 @@ def beanstalk_env_from_registry(registry):
...
-def find_other_in_pair(element, pair):
-    """ Wrapper for compute_set_difference_one
-
-        :arg element (T): item to look for in pair
-        :arg pair (2-tuple of T): pair of things 'element' is in
-        :returns (T): item in pair that is not element
-        :raises: exception if types do not match or in compute_set_diferrence_one
-    """
-    return compute_set_difference_one(set(pair), {element})
+# def compute_set_difference_one(s1, s2):
```
netsettler, line 299: This is unused, as mentioned above.

src/encoded/util.py
```diff
@@ -335,8 +335,8 @@ def parameterized_delay_rerun(*args):
 delay_rerun = customized_delay_rerun(sleep_seconds=1)
 
 
-def utc_today_str():
-    return datetime.datetime.strftime(datetime.datetime.utcnow(), "%Y-%m-%d")
+# def utc_today_str():
```
netsettler, line 338: Available from dcicutils.

netsettler (commented):
I've made a few comments to help reviewers get through this. Please feel free to ask if other things are left unclear.

willronchetti (approved):
We should consider adding some dedicated documentation on this file

alexkb0009 (approved):
Changes in code sections that am familiar-ish with look good to go, didn't check out the unit tests.

PR #1593, opened 2022-02-03
Upgrade Webpack to Version 5

utku-ozturk:


package.json
```diff
@@ -95,9 +95,14 @@
     "string-replace-loader": "^1.0.0",
     "terser-webpack-plugin": "^4.2.3",
     "uglifyjs-webpack-plugin": "^2.2.0",
-    "webpack": "4.45.0",
+    "webpack": "5.1.3",
     "webpack-bundle-analyzer": "^4.5.0",
-    "webpack-deadcode-plugin": "^0.1.14"
+    "webpack-deadcode-plugin": "^0.1.15",
+    "process": "0.11.10",
```
alexkb0009, line 101: Some of these libraries haven't been updated in 5+ yrs (e.g. 'process'), not sure if is good idea to include. Will try to play around with 'fallback' config a bit.

package.json
```diff
@@ -134,6 +139,10 @@
     "react-transition-group": "^4.4.1",
     "redux": "^3.7.0",
     "subprocess-middleware": "0.1.0",
-    "underscore": "^1.13.2"
```
alexkb0009, line 137: Should keep this at 10.13.2 (or higher)?

package.json
alexkb0009: Also if including in "dependencies" then don't need to replicate the entry in "devDependencies".

alexkb0009: Yeah seems like we need at least some polyfills for Webpack 5.
SPC currently requires 'crypto-js' library already, so will test on CGAP if can use that in place of crypto-browserify per e.g. [this answer](https://stackoverflow.com/questions/67572355/webpack-5-angular-polyfill-for-node-js-crypto-js#answer-67808287).

alexkb0009: Rest of this looks mostly good, would check if can exclude buffer and process though.
And make sure HiGlass works.
On CGAP-side might use different set of polyfills but still exploring..

alexkb0009: On CGAP-side, have found that many of these libraries aren't explicitly needed, but I'm unsure yet (maybe some feature deep in HiGlass gets broken or affected). For this reason am thinking we could maybe wait until HiGlass developers update their code to work with Webpack 5 and this would make efforts for us much simpler as we'd know HiGlass is up to date and not have to worry about adding our own polyfills for it.

alexkb0009: For crypto as well, for now just set "crypto": false instead of using a polyfill. Login seems to work at least, not sure if anything else been affected..

package.json
utku-ozturk: @alexkb0009 Yeah, we already upgraded it in master after the dependabot alert. I merged it into the webpack5 branch now.

alexkb0009 (approved):
Looks good, might consider removing 'buffer' and 'process' from package-json to see if that works, since those 2 libraries (especially 'process' -- [the one obtained from NPM](https://www.npmjs.com/package/process), not the [one bundled with NodeJS standard library](https://nodejs.org/api/process.html))(I think path-browserify contains own buffer and process? Not sure)

PR #1595, opened 2022-02-08
Add organism calcprop to biosource

aschroed:
Add a calcprop to Biosource to link Organism item directly.

This get's around the issue of having datasets that are from a known organism for which we have an Item in the database but no Individual schema.

The current case is hamster data but making this change will allow us to more easily add datasets from critters that we don't have a genome for (and possible never will) and still allow faceting on their organism.

This uses a field in Biosource 'override_organism_name' that must match the 'name' property of an Organism Item to generate the linkTo if there is no Individual linked to the Biosource.

Those items with a linked Individual get the Biosource.organism linkTo via the indirect connection to the Organism that is linked to the Individual so don't need and should not have the override_organism_name property.

NOTE: this needs I think at least 2 UI changes in order for the 'Organism' property in the summary info on the RepSet page to display the scientific name properly.  Trello ticket here: https://trello.com/c/rUy5Rjfg/550-ui-changes-to-use-biosource-organism-calculated-property

~~NOTE: it may be possible to cut down on some of the embeds though maybe not as most of the ones that might be removed are used in the calcprop.~~ Believe this has been done

### DIFF
--- a/src/encoded/schemas/biosource.json
+++ b/src/encoded/schemas/biosource.json
@@ -130,6 +130,15 @@
             "description": "A override name for the biosource that can be used as the biosource_name in place of the automatically generated one",
             "type": "string",
             "lookup": 1000
+        },
+        "override_organism_name": {
+            "title": "Organism Name Override",
+            "description": "An override name for the organism that can be used to find an Organism Item if the Biosource is not linked to an Individual",
+            "comment": "To get all the links to work this must be equivalent to the organism.name field",
+            "type": "string",
+            "exclude_from" : ["submit4dn"],
+            "permission": "import_items",
+            "lookup": 1000
         }
     },
     "facets": {
@@ -142,7 +151,7 @@
         "tissue.preferred_name": {
             "title": "Tissue"
         },
-        "individual.organism.name": {
+        "organism_name": {
             "title": "Organism"
         },
         "cell_line_tier": {
@@ -156,7 +165,7 @@
         }
     },
     "columns": {
-        "individual.organism.name": {
+        "organism_name": {
             "title": "Organism"
         },
         "biosource_type": {
--- a/src/encoded/schemas/biosource.json
+++ b/src/encoded/schemas/biosource.json
@@ -138,6 +138,9 @@
             "type": "string",
             "exclude_from" : ["submit4dn"],
             "permission": "import_items",
+            "suggested_enum": [
+                "hamster"
+            ],
             "lookup": 1000
         }
     },
--- a/src/encoded/schemas/biosource.json
+++ b/src/encoded/schemas/biosource.json
@@ -151,7 +151,7 @@
         "tissue.preferred_name": {
             "title": "Tissue"
         },
-        "organism_name": {
+        "organism.name": {
             "title": "Organism"
         },
         "cell_line_tier": {
@@ -165,7 +165,7 @@
         }
     },
     "columns": {
-        "organism_name": {
+        "organism.name": {
             "title": "Organism"
         },
         "biosource_type": {
--- a/src/encoded/types/experiment.py
+++ b/src/encoded/types/experiment.py
@@ -157,7 +157,6 @@ class Experiment(Item):
 
         # Organism linkTo
         'biosample.biosource.organism.name',
-        'biosample.biosource.individual.organism.name',
 
         # Modification linkTo
         'biosample.modifications.modification_type',
--- a/src/encoded/types/experiment.py
+++ b/src/encoded/types/experiment.py
@@ -157,8 +157,6 @@ class Experiment(Item):
 
         # Organism linkTo
         'biosample.biosource.organism.name',
-        'biosample.biosource.organism.scientific_name',  # not sure we need this one
-        'biosample.biosource.individual.organism.scientific_name',
         'biosample.biosource.individual.organism.name',
 
         # Modification linkTo
--- a/src/encoded/types/experiment.py
+++ b/src/encoded/types/experiment.py
@@ -156,8 +156,10 @@ class Experiment(Item):
         'biosample.biosource.individual.protected_data',
 
         # Organism linkTo
-        'biosample.biosource.individual.organism.name',
+        'biosample.biosource.organism.name',
+        'biosample.biosource.organism.scientific_name',  # not sure we need this one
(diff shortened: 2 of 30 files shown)


src/encoded/schemas/biosource.json
```diff
@@ -156,7 +165,7 @@
         }
     },
     "columns": {
-        "individual.organism.name": {
+        "organism_name": {
```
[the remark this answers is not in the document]
aschroed: Fixed that and one in the facets as well. Thanks for catching!

src/encoded/types/experiment.py
```diff
@@ -156,8 +156,10 @@ class Experiment(Item):
         'biosample.biosource.individual.protected_data',
 
         # Organism linkTo
-        'biosample.biosource.individual.organism.name',
+        'biosample.biosource.organism.name',
+        'biosample.biosource.organism.scientific_name',  # not sure we need this one
         'biosample.biosource.individual.organism.scientific_name',
```
sbreiff, line 161: Do we need this?

src/encoded/types/experiment_set.py
```diff
@@ -136,6 +136,7 @@ class ExperimentSet(Item):
         "experiments_in_set.biosample.biosource.biosource_type",
         "experiments_in_set.biosample.biosource.cell_line_tier",
         "experiments_in_set.biosample.biosource.override_biosource_name",
+        "experiments_in_set.biosample.biosource.override_organism_name",  # do we need this?
```
sbreiff, line 139: I think we do need this but it might be better placed under "Organism linkTo"

src/encoded/types/experiment_set.py
```diff
@@ -152,7 +153,10 @@ class ExperimentSet(Item):
         "experiments_in_set.biosample.biosource.tissue.synonyms",
 
         # Organism linkTo
+        "experiments_in_set.biosample.biosource.organism.name",  # calc prop
         "experiments_in_set.biosample.biosource.individual.organism.name",
+        "experiments_in_set.biosample.biosource.organism.scientific_name",
+        "experiments_in_set.biosample.biosource.individual.organism.scientific_name",
```
sbreiff, line 159: We probably won't need this one anymore

src/encoded/types/file.py
```diff
@@ -310,6 +310,8 @@ class File(Item):
         'experiments.biosample.biosource.individual.protected_data',
 
         # Organism linkTo
+        'experiments.biosample.biosource.organism.name',
+        'experiments.biosample.biosource.organism.scientific_name',
         'experiments.biosample.biosource.individual.organism.name',
         'experiments.biosample.biosource.individual.organism.scientific_name',
```
sbreiff, line 316: probably won't need this embed anymore

aschroed:
> I am not sure we will ever have a case where an individual is there, but we want to override the organism. Anyway, with this permission settings I think it's OK to allow us more flexibility since it's only us that can modify it.

I agree that this is unlikely and I also agree that this does provide flexibility - though some might argue that is not the best idea I feel it is safe enough and am OK with it as it takes admin to use as you point out.

src/encoded/types/experiment.py
```diff
@@ -156,8 +156,10 @@ class Experiment(Item):
...
+        'biosample.biosource.individual.organism.name',
```
sbreiff, line 162: this embed will also be redundant to `biosample.biosource.organism.name` so I think we can remove?

sbreiff (commented):
Overall the changes look good, but we should take a close look at the embeds so that we're not adding unnecessary/duplicate embeds.

The main example here are places where `...biosource.organism.name` is embedded as well as `...biosource.individual.organism.name`. If there is an Individual on the Biosource, then `biosource.individual.organism` is the same item as `biosource.organism` and I think embedding the `name` property twice is unnecessary. For the sake of invalidation scope, the concern is that if one item changes (e.g. the Organism item), we want to make sure all items that embed a relevant property are invalidated so they get updated, but embedding the same db item twice doesn't help invalidation as far as I can tell.

src/encoded/schemas/biosource.json
```diff
@@ -142,7 +151,7 @@
         "tissue.preferred_name": {
             "title": "Tissue"
         },
-        "individual.organism.name": {
+        "organism_name": {
```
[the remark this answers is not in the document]
aschroed: > Overall the changes look good, but we should take a close look at the embeds so that we're not adding unnecessary/duplicate embeds.
> 
> The main example here are places where `...biosource.organism.name` is embedded as well as `...biosource.individual.organism.name`. If there is an Individual on the Biosource, then `biosource.individual.organism` is the same item as `biosource.organism` and I think embedding the `name` property twice is unnecessary. For the sake of invalidation scope, the concern is that if one item changes (e.g. the Organism item), we want to make sure all items that embed a relevant property are invalidated so they get updated, but embedding the same db item twice doesn't help invalidation as far as I can tell.

@sbreiff I agree that this does need careful consideration as to not have unnecessary redundant embeds.  The reason that I left most if not all of the ones in question there was that almost always the calculated property uses the link from biosource to individual to organism to make the linkTo and therefore a change to the Individual item (by changing it's organism) should trigger invalidation of the containing items.  It was my understanding that to ensure appropriate invalidation of items based on calculated properties you needed to embed all the fields used in the calcprop - but it's been awhile so that may not be the case?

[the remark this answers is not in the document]
sbreiff: the path to the item doesn't matter, just the itemtype and property. so having biosource.individual.organism.name is equivalent to biosource.organism.name. Biosource is the item that has the calculated property, so biosource will need to embed both, but experiment and experiment_set and biosample do not.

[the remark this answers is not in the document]
aschroed: OK - got it.  I think I've trimmed unnecessary/redundant embeds.

aschroed:
UI changes look good to me.  I'll add @alexkb0009 as a reviewer and if signs off I think we should be good to go.

alexkb0009 (approved):
JS changes look good to me, didn't test/look-through-too-thoroughly but if everything is working as expected it should be good. I will check over the SPC edits and maybe make some comments there if needed (else approve).

PR #1602, opened 2022-02-17
Add biorxiv version in schemas

andreacosolo:


aschroed (approved):
Looks good.  The OCD part of me wants you to use single quotes in line 123 to be consistent but it's not worth changing for that reason.

PR #1604, opened 2022-02-22
Inserts cleanup

netsettler:
This is an old branch with a bunch of mostly-PEP8 inserts-related cleanups I did a long time ago and never checked in. I merged master so the rest of the repo is up-to-date, and there was only one conflict that I resolved in favor of master just to keep things simple. 

There are a small number of places where there were duplicated tags in a dictionary that had to be resolved. e.g.,  in src/encoded/tests/data/inserts/file_processed.json at line 223, the "notes" key is used twice, so the first is removed since I believe that preserves the existing behavior.

```
{
    "lab":"/labs/4dn-dcic-lab/",
    "award":"/awards/1U01CA200059-01/",
    "notes":"sample dcic notes",
    "status":"uploaded",
    "filename":"4DNFIN232ZZZ.pairs.gz",
    "file_format": "d13d06cf-218e-4f61-aaf0-91f226248b2c",
    "notes": "pairs format item",
    "uuid":"0048955c-7cb6-4e56-a4d8-56fad52f9999"
}
```
because In a python I see:
```
Python 3.6.9 (default, Dec  5 2019, 22:00:51) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.12)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> {"foo": "something", "foo": "something else"}
{'foo': 'something else'}
>>> 
```
The fact of _so many_ syntax errors makes it hard to get good diagnostics when there are real problems of this kind, which is why I'd like to repair these.


aschroed (approved):
Looks fine to me.

willronchetti (approved):

PR #1606, opened 2022-02-23
Experiment Set Item Page - Biosource Updates

utku-ozturk:
Trello: https://trello.com/c/r4XElyp9

On the ExperimentSetReplicate page can we make the following changes to the Experiment Set Properties section:
Biosource Type:
   Rename to 'Sample Type'
   Get value from experiments_in_set.biosample.biosample_type

Biosource:
   Rename to 'Sample'
   No change to where the text to display comes from.
   Make this a link to the experiments_in_set.biosample.biosource page
    or if more than one different Biosource then to a search result page of those biosources.

Screen Shot 2022-02-23 at 14 32 05


alexkb0009 (approved):
Looks great!

PR #1615, opened 2022-03-03
Default diff

andreacosolo:
Many item types have properties which are modified by the item's `_update()` method. These properties **might** change every time the item is patched, so they should always be included in the `default_diff` for proper invalidation scope computation.

However, in practice this is true primarily for Publication and Gene, which get metadata from external resources via `_update`. For all other item types, `_update` works in a similar way to calculated properties: it changes a property based on the value of other properties. Therefore, it makes more sense to adjust the embed list, instead of adding properties to `default_diff`, to avoid unnecessary invalidation.

Consider the following example: in OntologyTerm items, `preferred_name` is updated based on `term_name`. Therefore, instead of adding `preferred_name` to the default_diff (which would **always** invalidate linked items embedding it, regardless of which property of OntologyTerm is modified), it is sufficient to make sure that also `term_name` is included in the embeds for OntologyTerm (only modifications to `term_name` will cause invalidation of linked items that embed `preferred_name`). 

Note 1: default_diff for Publication was added previously #1598 
Note 2: this PR ignored AccessKey, Biosample, File, User.


aschroed (approved):
This looks good to me.  I notice that at least one of the quality_metric items bamcheck does not appear to be used in fourfront (though it is in CGAP) - but no reason not to leave the default_diff.

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "3.6.3"
+version = "3.6.4"
```
willronchetti, line 4: Will need to be `3.6.7` at least now

willronchetti (approved):
Changes look good given the various `_update` methods

PR #1621, opened 2022-03-16
Version Updates: Node v16, React v17

utku-ozturk:
Trello: https://trello.com/c/cvHNgj9r


alexkb0009:
I'll make a few small edits then push + merge all Node16 branches (will update this [line](https://github.com/4dn-dcic/fourfront/blob/master/.ebextensions/20_packages.config#L18) to v16, remove 'node' entry from package.json, and checking/testing)

PR #1626, opened 2022-03-29
Jest Updates

utku-ozturk:
Jest tests are failing in both the local environment and GH NPM runs after node16 and react17 transition.


src/encoded/static/components/app.js
```diff
@@ -790,7 +791,7 @@ export default class App extends React.PureComponent {
                     window.location.replace(targetHref + hashAppendage);
                 } else {
                     const [ old_path ] = ('' + window.location).split('#');
-                    window.location.assign(targetHref + hashAppendage);
+                    Sinon.stub(targetHref + hashAppendage);
```
alexkb0009, line 794: [Sinon.stub](https://sinonjs.org/releases/latest/stubs/) (and Sinon in general) is meant to be used within unit tests, not within the app code itself, I'm pretty sure..

PR #1629, opened 2022-04-06
Add wf engine field to schema

clarabakker:
Adds two fields (workflow_engine and run_args) to the workflow.json schema for the purpose of using caper in automated workflow runs.


aschroed (approved):
Don't forget to bump the version in pyproject.toml 
Failing tests are due to front end incompatibilites that Utku and Ugur are working on fixing so they are fine to ignore

PR #1637, opened 2022-04-21
Mobile UX Updates

utku-ozturk:
Trello: https://trello.com/c/sgW1HET

Homepage barchart ipad
Screen Shot 2022-04-28 at 16 23 04

Homepage barchart iphone
Screen Shot 2022-04-28 at 16 22 30


alexkb0009:
Glanced at code and looks good from what can tell. Am interested in how the server-side detection of mobile browser works in long run.. it might not be too stable since user-agents and browsers can evolve over time; another idea might be to always assume is mobile content for initial server-side render, and then after client-side render, figure out isMobile based on `props.windowWidth`. Would 'detect' that is client-side render based on state.mounted. Maybe something like: 

```javascript
 componentDidMount(){
     this.setState({ mounted:true })
 }
 render(){
     const { windowWidth, ... } = this.props;
     const isMobile = !state.mounted ? true : (windowWidth < 784);
     const initialFields = ... 
.. }`

PR #1638, opened 2022-04-21
Repair nginx.conf

willronchetti:
- Do not 301 to HTTPS on staging

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "4.1.0"  # 4.0.0 introduced containerization
+version = "4.1.1"  # 4.0.0 introduces containerization
```
netsettler, line 4: Should continue to use past tense here, or just remove the comment.

netsettler (approved):
Comment is not blocking, but it would be nice to fix.

PR #1639, opened 2022-04-25
Repair Fargate OOM

willronchetti:
- Repairs `memlimit` memory recycling mechanism by eliminating reliance on `psutil.process.memory_percent`, which does not report accurate values on Fargate
- Wrap fourfront API workers with `supervisord`, since `memlimit` will kill the waitress processes once they exceed 400Mb
- Provide a dedicated backup server for when workers get killed due to OOM


src/encoded/__init__.py
```diff
@@ -27,6 +28,7 @@
 
 # location of environment variables on elasticbeanstalk
 BEANSTALK_ENV_PATH = "/opt/python/current/env"
+FOURFRONT_ECS_REGION = 'us-east-1'
```
netsettler, line 31: Note that `dcicutils.common.REGION` knows this value.

netsettler (approved):
This looks fine. My comment is not blocking.

PR #1661, opened 2022-06-10
June Security Update

willronchetti:
- Removes Python 3.6 support
- Builds Docker under Python 3.8, resulting in numerous library updates closing various vulnerabilities 
- Brings in upgrader fix from snovault
- Adds test from CGAP verifying schema_version presence
- Ensure `/auth0_config` API returns needed results

### DIFF
--- a/src/encoded/tests/test_schemas.py
+++ b/src/encoded/tests/test_schemas.py
@@ -197,3 +197,21 @@ def test_fourfront_crawl_schemas(testapp, registry):
     field_schema = crawl_schema(registry[TYPES], field_path, schema)
     assert isinstance(field_schema, dict)
     assert field_schema['title'] == 'File Size'
+
+
+def test_schema_version_present_on_items(app):
+    """Test a valid schema version is present on all non-test item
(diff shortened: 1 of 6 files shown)


src/encoded/tests/test_schemas.py
```diff
@@ -197,3 +197,21 @@ def test_fourfront_crawl_schemas(testapp, registry):
...
+    all_types = app.registry.get(TYPES).by_item_type
+    for type_name, item_type in all_types.items():
+        if type_name.startswith("testing"):
+            continue
+        schema_version = item_type.schema_version
+        if item_type.is_abstract is False:
+            assert schema_version
+            assert int(schema_version) >= 1
+        else:
+            assert schema_version == ""
```
netsettler, line 217:
```suggestion
        if item_type.is_abstract:
            assert schema_version == ""
        else:
            assert schema_version
            assert int(schema_version) >= 1
```

netsettler (approved):
My comment is not blocking.

PR #1666, opened 2022-06-15
WF Node QC Metrics Table

utku-ozturk:
Trello: https://trello.com/c/BOFznhhl

image

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "4.3.2"  # 4.0.0 introduced containerization
+version = "4.3.3"  # 4.0.0 introduced containerization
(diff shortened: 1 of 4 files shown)


alexkb0009:
Looks good, if have any ideas about simplifying MetricsView or improving it (or any of these views) then go for it also.

I'm reading the old code I wrote here and being like.. what the heck was I thinking, why am I merging fields into other items, haha.

alexkb0009 (approved):

utku-ozturk:
> Looks good, if have any ideas about simplifying MetricsView or improving it (or any of these views) then go for it also.
> 
> I'm reading the old code I wrote here and being like.. what the heck was I thinking, why am I merging fields into other items, haha.

I am not sure but QC schemas were used to include only a few of the metrics explicitly until we defined them with Soo 2 years ago. So the table was more concise and displays ~3  properties and used to work properly.

PR #1669, opened 2022-06-16
Fix - SPC Symlink Prevents Authorization

utku-ozturk:
Trello: https://trello.com/c/ah5CqjTQ

When the fourfront project is linked with the shared portal, it gets auth0-lock error. Unable to login.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -92,14 +92,16 @@ const resolve = {
     }
 };
 
-// // Common alias, hopefully is fix for duplicate versions of React
-// // on npm version 7+ and can supersede `./setup-npm-links-for-local-development.js`.
-// // @see https://blog.maximeheckel.com/posts/duplicate-dependencies-npm-link/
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -98,7 +98,9 @@ const resolve = {
 spcPackageJson = require("@hms-dbmi-bgm/shared-portal-components/package.json");
 spcPeerDependencies = spcPackageJson.peerDependencies || {};
 Object.keys(spcPeerDependencies).forEach(function(packageName) {
-    resolve.alias[packageName] = path.resolve("./node_modules/" + packageName);
+    if (packageName !== 'auth0-lock') {
+        resolve.alias[packageName] = path.resolve("./node_modules/" + packageName);
+    }
```
alexkb0009, line 103: Looks good, am curious if now that we're on React 17 we can get rid of this Object.keys(...).foreEach(...) block entirely (not sure I tested it without this on React17), but not too important / very low priority.

PR #1706, opened 2022-08-05
Syntax makeover for clear-db-es-contents

netsettler:
* Port some argument changes to clear-db-es-contents from `cgap-portal`.
* Create a `.flake8` file.

### DIFF
--- a/src/encoded/tests/test_clear_db_es_contents.py
+++ b/src/encoded/tests/test_clear_db_es_contents.py
@@ -1,14 +1,48 @@
+import contextlib
 import pytest
 
+from dcicutils.common import OrchestratedApp
+from dcicutils.env_utils import EnvUtils, APP_FOURFRONT
+from dcicutils.lang_utils import disjoined_list
 from unittest import mock
+from ..commands import clear_db_es_contents as clear_db_es_contents_module
 from ..commands.clear_db_es_contents import (
     clear_db_tables,
-    run_clear_db_es
+    run_clear_db_es,
+    main as clear_db_es_contents_main
 )
-from ..commands import clear_db_es_contents
 
 
-pytestmark = [pytest.mark.setone, pytest.mark.working]
+pytestmark = [pytest.mark.setone, pytest.mark.working, pytest.mark.indexing]
+
+
+def is_running_app(app: OrchestratedApp) -> bool:  # Move to dcicutils.env_utils
+    return app == EnvUtils.ORCHESTRATED_APP
+
+
+class MockLog:
+
+    def __init__(self, *, messages=None, key=None):
+        self.messages = messages or {}
+        self.key = key
+
+    def _addmsg(self, kind, msg):
+        self.messages[kind] = msgs = self.messages.get(kind, [])
+        msgs.append(msg[self.key] if self.key else msg)
+
+    def error(self, msg):
+        self._addmsg('error', msg)
+
+    def info(self, msg):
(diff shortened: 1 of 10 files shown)


src/encoded/tests/test_clear_db_es_contents.py
```diff
@@ -22,56 +56,254 @@ def test_clear_db_tables(app, testapp):
...
 
-def test_run_clear_db_envs(app, testapp):
+_FOURFRONT_PRODUCTION_ENVS = ['fourfront-production-blue', 'fourfront-production-green', 'data', 'staging']
+# Really we only care about the first of these names, but the rest are names that were at one time
+# planned to be stg or prd names for cgap, so we'll use them to tell that run_clear_db_es is properly
+# skipping any such names. -kmp 4-Jun-2022
+_CGAP_PRODUCTION_ENVS = ['fourfront-cgap', 'fourfront-cgap-green', 'cgap-green', 'fourfront-cgap-blue', 'cgap-blue']
 
-    orig_env = app.registry.settings.get('env.name')
+_PRODUCTION_ENVS = _FOURFRONT_PRODUCTION_ENVS if is_running_app(APP_FOURFRONT) else _CGAP_PRODUCTION_ENVS
```
willronchetti, line 65: Probably don't need to do this sort of thing and can just assign the FF names directly.

src/encoded/tests/test_clear_db_es_contents.py
```diff
@@ -22,56 +56,254 @@ def test_clear_db_tables(app, testapp):
...
+                    assert mock_run_create_mapping.call_count == expected_es_clears
+
+                    allowed_envs = ['fourfront-test-env-decoy-1', 'fourfront-test-env-decoy-2']
+                    # test if we are only running on specific envs
+                    with logged_messages(error=[(f'clear_db_es_contents: The current environment, {test_env},'
+                                                 f' is not {disjoined_list(allowed_envs)}.'
+                                                 f' Skipping the attempt to clear DB.')]):
+                        assert run_clear_db_es(app, only_envs=allowed_envs, skip_es=False) is False
+                    expected_db_clears += 0
+                    expected_es_clears += 0
```
willronchetti, line 200: Same comment from CGAP - I get being explicit here but still feels unnecessary.

willronchetti (approved):

src/encoded/tests/test_clear_db_es_contents.py
netsettler: I want to share the script between both repos.

src/encoded/tests/test_clear_db_es_contents.py
netsettler: If I didn't do this, I'd put the same thing commented out. It seems harmless, so I'm ignoring your uneasiness. :)

src/encoded/tests/test_clear_db_es_contents.py
netsettler: I have a plan in the back of my mind to pull these things out to a common repo that is actually shared.

src/encoded/tests/test_clear_db_es_contents.py
netsettler: The idea is so visually as you scan down the page, you can read off the number of expected changes for each section. The math is secondary to the self-documentation.

netsettler:
This is passing tests locally on my machine at this point.

PR #1707, opened 2022-08-11
Repair local deploys

willronchetti:
- Disabled mpindexer, which is not used in production and does not respect ini file settings
- Disabled repoze.debug egg pipeline
- Pass `GLOBAL_ENV_BUCKET` to docker local
- Document setting `GLOBAL_ENV_BUCKET` in `docker-local.rst`
- Update documentation so readthedocs links to Docker documentation


docker-compose.yml.template
```diff
@@ -50,6 +50,8 @@ services:
       # Default to caller's environment
       AWS_ACCESS_KEY_ID:
       AWS_SECRET_ACCESS_KEY:
+      # Needed to resolve env info
+      GLOBAL_ENV_BUCKET:
```
netsettler, line 54: Why no value? Shouldn't it be the cgap devtest global bucket env, `cgap-devtest-main-foursight-envs`?

netsettler (approved):

docs/source/docker-local.rst
```diff
@@ -38,7 +38,11 @@ Use the ``prepare-docker`` command to configure ``docker-compose.yml`` and ``doc
     --run-tests           if supplied, causes tests to be run in container
                         (default: not tested)
     --s3-encrypt-key-id S3_ENCRYPT_KEY_ID
-                        an encrypt key id (default: the empty string)
+                        an encrypt key id (default: the empty string), not typically used for FF
+
+
+Note that you must additionally set ``GLOBAL_ENV_BUCKET=foursight-prod-envs``, which will be passed to
+the container to resolve environment information.
```
netsettler, line 46: Should also add (form cgap-portal):

```
On initial run, you will want to run with the ``--load-inserts`` option so data is loaded.
Pass ``--data-set local`` to get local inserts, or ``deploy`` to use the production inserts.
```

PR #1720, opened 2022-09-06
Fix build problem caused by wheel version incompatibility (C4-906)

netsettler:
* Pin `poetry` version in `Makefile` to `1.1.15`
* Pin `wheel` in `pyproject.toml` to `0.37.1`
* Update `poetry.lock` for changes to `flake8` and `wheel`. (The `flake8` update is because we needed to pick up a newer version, not because we needed to change `pyproject.toml`.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,12 +33,18 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 # Note that the ordering of these operations is intentional to minimize package footprint
 WORKDIR /home/nginx/.nvm
 ENV NVM_DIR=/home/nginx/.nvm
-COPY deploy/docker/production/install_nginx.sh /
+COPY deploy/docker/production/install_nginx.sh /install_nginx.sh
+
+# Temporarily replacing
+#     curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/venv python - && \
+# with
+#     pip install poetry==1.1.15
+# because of a problem with wheel compatibility in 1.2.0. Need to debug that later. -kmp 31-Aug-2022
 RUN apt-get update && apt-get upgrade -y && \
     apt-get install -y --no-install-recommends vim emacs net-tools ca-certificates build-essential \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -33,12 +33,18 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
...
+#     curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/venv python - && \
+# with
+#     pip install poetry==1.1.15
+# because of a problem with wheel compatibility in 1.2.0. Need to debug that later. -kmp 31-Aug-2022
 RUN apt-get update && apt-get upgrade -y && \
     apt-get install -y --no-install-recommends vim emacs net-tools ca-certificates build-essential \
     gcc zlib1g-dev postgresql-client libpq-dev git make curl libmagic-dev && \
     pip install --upgrade pip && \
-    curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/venv python - && \
+    pip install poetry==1.1.15 && \
```
willronchetti, line 47: I would consider pinning this via env variable similar to others above, like nvm/node version

pyproject.toml
```diff
@@ -149,7 +149,7 @@ pytest-timeout = ">=1.0.0"
 pytest-xdist = ">=1.14"
 "repoze.debug" = ">=1.0.2"
 
-wheel = ">=0.29.0"  # needed for distribution but not any particular version
+wheel = "0.37.1"
```
willronchetti, line 152: Does this have any real impact since it is in dev-dependencies?

willronchetti (approved):

pyproject.toml
netsettler: Well, even in dev dependencies, the solver will complain if there is no way to reconcile the set of claimed versions. It wants to resolve this as lock time, not at load time.

Dockerfile
netsettler: I'll do that in a later PR. I'll queue a ticket for that. For now I just want it to match what we're doing in cgap-portal, and I didn't do it that way there.

PR #1725, opened 2022-09-07
Clean NPM Cache in Docker

willronchetti:
- Adds a cache clean after `npm ci` to reduce image size


netsettler (approved):

willronchetti:
There is a .cache folder that npm uses and duplicates the node_modules unnecessarily, I realized it when reviewing the ECR scans and the cache paths showed vulnerabilities

PR #1727, opened 2022-09-10
Manage development.ini and test.ini outside of source control

netsettler:
This PR is conceptually identical to [cgap-portal PR 637](https://github.com/dbmi-bgm/cgap-portal/pull/637) but with the fix already incorporated that we later ended up needing as [cgap-portal PR 642](https://github.com/dbmi-bgm/cgap-portal/pull/642).

This change is intended only to affect developers who are doing local testing (e.g., `make test` or a call to `pytest`) that would use `test.ini` or who are doing local deploys (e.g., `make deploy1`) that would use `development.ini`.

This is technically a breaking change, in that it may affect developer scripts, but **this should not affect production builds or GA, so we've agreed to only bump the minor version, not the major version.** Nevertheless, you should report problems in production, GA, or docker management if you see them.** 

Prior to this change, `development.ini` and `test.ini` were in source control. This PR changes this so that what's in source control is `development.ini.template` and `test.ini.template`. There is a command introduced, `prepare-local-dev` that you can run to create a `development.ini` and `test.ini`. Once the file exists, the `prepare-local-dev` command will not touch it, so you can do other edits as well without concern that they will get checked in. The primary change that this command does is to make a local environment of `fourfront-devlocal-<yourusername>` or `fourfront-test-<yourusername>` so that testing and debugging that you do locally will be in an environment that does not collide with other users. You can give a `--env` argument to `prepare-local-dev` to use a different name, though it's probably easier to just edit the resulting file.

Note to reviewers (especially @willronchetti): The `test.ini` case is a little weird because some things in that file are ignored. We might want to make it so that `conftest_settings.py` somehow pulls data from there or is consistent with there or figure out why they diverge and how to describe when to use one or the other.


Makefile
```diff
@@ -88,6 +88,7 @@ build-after-poetry:  # continuation of build after poetry install
 	make npm-setup-if-needed
 	poetry run python setup_eb.py develop
 	make fix-dist-info
+	poetry run prepare-local-dev
```
willronchetti, line 91: You might want to ensure inserts are loaded when this is run

willronchetti (approved):

Makefile
netsettler: I'm not sure I understand why inserts are relevant to this. This is just about setting up details that need to be user-specific, like the indexing prefix. Can you elaborate on your concern?

Yes, this is a very old PR. I was surprised it was never merged, but it fortunately seems not to have gotten stale, so I'm trying to revive it.

netsettler:
Local testing results:
```
=== 1568 passed, 9 skipped, 209 deselected, 706 warnings in 1380.02s (0:23:00) ===
=== 184 passed, 3 skipped, 1599 deselected, 66 warnings in 285.15s (0:04:45) ===
commit d53c6dc226fec3c0f36d80682c5cc9ff2a8f0e7c (HEAD -> kmp_custom_ini, origin/kmp_custom_ini)
Mon Feb 20 22:10:48 EST 2023
```

netsettler:
I've also deployed this branch to `fourfront-mastertest` successfully.
So I'm going to go ahead and merge.

PR #1729, opened 2022-09-15
Get sentry DSN from GAC, and do other initialization leveling with cgap-portal (C4-913)

netsettler:
* Address [Fourfront Sentry DSN needs to be managed via Secrets Manager (C4-913)](https://hms-dbmi.atlassian.net/browse/C4-913)
* Some other leveling between Fourfront and CGAP versions of `src/encoded/__init__.py`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@ fourfront
 Change Log
 ----------
 
+4.5.19
+======
+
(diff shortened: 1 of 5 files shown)


willronchetti (approved):

netsettler:
This passed the full test suite on my local machine.

PR #1732, opened 2022-09-27
ElasticSearch 7 Support

willronchetti:
- Adds support for ES7
- Integrates new SQLAlchemy version
- Repairs broken test segmentation (should reduce test time)
- Add `?skip_indexing` parameter
- Adds B-Tree index on max_sid to optimize indexing


docs/source/index.rst
```diff
@@ -65,7 +65,7 @@ NOTES:
 * If you try to invoke elasticsearch and it is not found,
   you may need to link the brew-installed elasticsearch::
 
-    $ brew link --force elasticsearch@5.6
+    $ brew link --force opensearch
```
netsettler, line 68: no version?

pyproject.toml
```diff
@@ -89,13 +89,13 @@ semantic_version = ">=2.9.0"
 sentry-sdk = "^1.5.6"
 simplejson = "^3.17.0"
 SPARQLWrapper = "^1.8.5"
-SQLAlchemy = "1.3.16"  # Pinned because >=1.3.17 is broken for us (circular constraints prevent deletes)
+SQLAlchemy = "1.4.41"  # pinned for stability
```
netsettler, line 92: say why that version? is that the latest as of today or something we've already tested or something else?

pyproject.toml
```diff
@@ -141,9 +141,8 @@ moto = "1.3.7"
 pip-licenses = "^3.5.1"
 pipdeptree = "^2.1.0"
 # TODO: Investigate whether a major version upgrade is allowable for 'pytest', which is several versions behind.
-pytest = ">=3.10,<4"
+pytest = "^5.0.0"
 pytest-cov = ">=2.2.1"
```
netsettler, line 145: I would have just commented-out pytest-exact-fixtures for a while so we have some temporary memory that it was there in case a question arises later. can always dig in git, but i find that a hassle. YMMV.

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "4.7.7"
+version = "5.0.0"
```
netsettler, line 4: did you already run a beta?

src/encoded/commands/clear_db_es_contents.py
```diff
@@ -37,9 +38,10 @@ def clear_db_tables(app):
     meta.reflect()
     connection = session.connection().connect()
     try:
-        # truncate tables by only deleting contents
-        for table in meta.sorted_tables:
-            connection.execute(table.delete())
+        # truncate tables by only deleting contents (sqlalchemy 1.4+ compliant)
+        table_names = ','.join(table.name for table in reversed(Base.metadata.sorted_tables))
+        connection.execute('SET statement_timeout = 300000;')  # give 5 mins for DB clear
+        connection.execute(f'TRUNCATE {table_names} RESTART IDENTITY;')
```
netsettler, line 44: 1. Is there any need for a CASCADE argument here?
2. should you be doing this as a single transaction so that it doesn't try to do conflict detection until you're done?
Maybe neither of these matter but I wanted to ask the question. I guess if you're getting OK results it probably doesn't matter. 

(Personally, I'm inclined to think you should DROP the DB and copy in a fresh empty DB from elsewhere rather than even doing these actions. I think actions like these are all risks of pointless inconsistency detection or even of loss of time proving that won't happen. I think copying a new DB into place of an old one is likely to be _way_ faster than this. It would be OK to just insert a comment suggesting this be explored in the future.)

Makefile
```diff
@@ -145,20 +145,20 @@ test-any:
 
 
 test-npm:
-	bin/test -vv --timeout=300 -m "working and not manual and not integratedx and not performance and not broken and not sloppy and workbook"
+	bin/test -xvv --durations=50 --timeout=300 -m "working and not manual and not integratedx and not performance and not broken and not sloppy and workbook"
```
netsettler, line 148: nice if this works. :)

src/encoded/search.py
```diff
@@ -352,13 +352,13 @@ def get_all_subsequent_results(initial_search_result, search, extra_requests_nee
             yield hit
 
 def execute_search_for_all_results(search):
-    size_increment = 100 # Decrease this to like 5 or 10 to test.
+    size_increment = 100  # Decrease this to like 5 or 10 to test.
```
netsettler, line 355: I know you didn't create this variable, but while we're in here: I have a mild preference for the variable name `chunk_size` for this. You're not incrementing something, you're just dividing it into chunks.

src/encoded/tests/conftest.py
```diff
@@ -38,8 +75,26 @@ def app_settings(request, wsgi_server_host_port, conn, DBSession):  # noQA - We
...
-# INDEXER_NAMESPACE_FOR_TESTING is defined in src/encoded/tests/workbook_fixtures.py
-# (In CGAP it would be defined here, and maybe in the future it will be for Fourfront, too.)
+INDEXER_NAMESPACE_FOR_TESTING = generate_indexer_namespace_for_testing('cgap')
+
+
+@pytest.fixture(scope='session')
+def es_app_settings(wsgi_server_host_port, elasticsearch_server, postgresql_server, aws_auth):
+    settings = make_app_settings_dictionary()
+    settings['create_tables'] = True
+    settings['persona.audiences'] = 'http://%s:%s' % wsgi_server_host_port  # 2-tuple such as: ('localhost', '5000')
```
netsettler, line 85: Make sure I didn't make a typo, but maybe:
```suggestion
    host, port = wsgi_server_host_port  # 2-tuple such as: ('localhost', '5000')
    settings['persona.audiences'] = f'http://{host}:{port}'
```

netsettler (approved):
I didn't see anything blocking.

src/encoded/tests/conftest.py
```diff
@@ -38,8 +76,26 @@ def app_settings(request, wsgi_server_host_port, conn, DBSession):  # noQA - We
     return settings
 
 
-# INDEXER_NAMESPACE_FOR_TESTING is defined in src/encoded/tests/workbook_fixtures.py
-# (In CGAP it would be defined here, and maybe in the future it will be for Fourfront, too.)
+INDEXER_NAMESPACE_FOR_TESTING = generate_indexer_namespace_for_testing('cgap')
+
```
aschroed, line 80: Is 'cgap' intended here?

aschroed (approved):
I had a brief look.  I see node16 is used.  I assume that this is all cool with UI team?  Also I noted one use of 'cgap' in tests which may have been intentional but I wasn't sure so added a comment.  From a brief scan all looks like good stuff to me.
Note that I believe Gulsah is deploying some of her UI branches to mastertest so might want to make sure she knows about planned updates and deployments.

PR #1739, opened 2022-10-13
Restore ca-certificates

willronchetti:
- Do not purge ca-certificates package, it is required for mirroring/JH to work properly (cannot call out via HTTPS without it)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,8 +85,7 @@ RUN poetry install --no-dev -vvv && \
 ENV NODE_ENV=production
 RUN npm run build && \
     npm run build-scss && \
-    rm -rf node_modules/ && \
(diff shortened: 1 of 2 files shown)


dmichaels-harvard (approved):

Dockerfile
```diff
@@ -85,8 +85,7 @@ RUN poetry install --no-dev -vvv && \
 ENV NODE_ENV=production
 RUN npm run build && \
     npm run build-scss && \
-    rm -rf node_modules/ && \
-    apt-get remove --purge --auto-remove -y ca-certificates
+    rm -rf node_modules/
```
netsettler, line 88: Maybe also add a comment saying that you used to remove ca-certificates but no longer do and explain why so no one tries the experiment again?

PR #1742, opened 2022-10-26
Fix @@download for extra_files

willronchetti:
- Implements file location checking in `open_data_url` and `build_external_creds` to ensure we get the file locations correct

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,7 @@ xlwt = "1.2.0"
 "zope.sqlalchemy" = "1.3"
 codeguru-profiler-agent = "^1.2.4"
 supervisor = "^4.2.4"
+wheel = "0.37.1"
 
 
 [tool.poetry.dev-dependencies]
@@ -148,8 +149,6 @@ pytest-mock = ">=0.11.0"
 pytest-timeout = ">=1.0.0"
 pytest-xdist = ">=1.14"
 "repoze.debug" = ">=1.0.2"
-
-wheel = "0.37.1"
 
 
 # In pytest 6.0, we'll be able to use this instead of a separate pytest.ini configuration.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "4.5.26"
+version = "4.6.0"
 description = "4DN-DCIC Fourfront"
 authors = ["4DN-DCIC Team <support@4dnucleome.org>"]
 license = "MIT"
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -836,7 +836,7 @@ class File(Item):
         return location
 
     def get_open_data_url_or_presigned_url_location(self, external, request, filename, datastore_is_database) -> str:
-        """  Returns the Open Data S3 url for the file if present (as a calculated property), and otherwise returns
+        """ Returns the Open Data S3 url for the file if present (as a calculated property), and otherwise returns
             a presigned S3 URL to a 4DN bucket. """
         open_data_url = None
         if datastore_is_database:  # view model came from DB - must compute calc prop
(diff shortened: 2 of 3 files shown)


aschroed (approved):
Just to make sure I understand - as long as external creds are built correctly then @@download will work fine.  And you've updated the build_external_creds function to have location info for where the actual file lives.

The change to _open_data_url was not strictly necessary - was it?  Just made this more efficient as the majority of files will be in the wfoutput bucket - or am I missing something.

willronchetti:
Correct, by fixing external_creds I believe it will be fixed downstream (will test on staging before swap). I think the change open_data_url is necessary just in case we decide to have the opposite mismatch case in the future - right now we are seeing `reference_file` with `processed_file` as an extra file but we could also see `processed_file` with a `reference_file` as an extra file. The change ensures that is handled correctly as well in case we do decide to do it, but you're right at this time at least it should never get there as they will always pass the first check under `wfoutput`.

pyproject.toml
```diff
@@ -149,8 +150,6 @@ pytest-timeout = ">=1.0.0"
 pytest-xdist = ">=1.14"
 "repoze.debug" = ">=1.0.2"
```
netsettler, line 152: if this goes away, it may need some compensating work in some make situation because:
* `wheel` is not installed by default
* `wheel` does not show up in `pip freeze` even if installed, so it's hard to see that it's the key needed component (note that it does appear in `pip list` in this case)
* I _think_ the absence of `wheel` means some operations with github fail to included essential components and i think results in our thinking something is not available when it is

src/encoded/types/file.py
```diff
@@ -866,25 +866,28 @@ def _open_data_url(self, status, filename):
...
                 # Returning a valid S3 URL to the public url of the file
-                self._head_s3(client, open_data_bucket, open_data_key)
+                try:
+                    self._head_s3(client, open_data_bucket, open_data_key)
+                except ClientError:
+                    continue  # try the other key
                 location = 'https://{open_data_bucket}.s3.amazonaws.com/{open_data_key}'.format(
-                    open_data_bucket=open_data_bucket, open_data_key=open_data_key,
+                    open_data_bucket=open_data_bucket, open_data_key=key,
                 )
```
netsettler, line 887: I think (untested):
```
                location = f"https://{open_data_bucket}.s3.amazonaws.com/{key}"
```

netsettler: (I would also be OK renaming the iteration variable from `key` to `some_open_data_key` if you wanted to remind yourself that it was in fact such a key. But that's a personal taste thing.)

src/encoded/types/file.py
```diff
@@ -919,8 +922,32 @@ def build_key(cls, registry, uuid, properties):
 
     @classmethod
     def build_external_creds(cls, registry, uuid, properties):
```
netsettler, line 924:
```suggestion
    def build_external_creds(cls, registry, uuid, properties, try_buckets=None):
        if try_buckets is None:
            try_buckets = [registry.settings['file_wfout_bucket'], registry.settings['file_upload_bucket']]
```
so later you can iterate over those. I think it will make it easier to test and possibly more reusable later.

src/encoded/types/file.py
```diff
@@ -919,8 +922,32 @@ def build_key(cls, registry, uuid, properties):
...
+
+            We could also be doing an upload, in which case we will do two head requests
+            that will give us nothing and we should fallback to the bucket associated
+            with the type.
+        """
+        conn = boto3.client('s3')
+        bucket = None
         key = cls.build_key(registry, uuid, properties)
+        # _head_s3 for both files and wfoutput buckets if we are doing a download
+        for b in [registry.settings['file_wfout_bucket'], registry.settings['file_upload_bucket']]:
```
netsettler, line 941: If you make the above-proposed change to add an extra parameter, this (untested) code might be better:
```suggestion
        for b in try_buckets:
```

netsettler: I wouldn't mind seeing `b` renamed to `try_bucket` for clarity.

pyproject.toml
netsettler: Oh, never mind. As you explained in Slack, @willronchetti, this dependency only got moved. There may still be an issue where it's loading the wrong version, but not due to this PR.

netsettler (approved):
No blockers here.

PR #1746, opened 2022-11-14
pi_name calculated property

aschroed:
Added a calculated property 'pi_name' that grabs the display_name from the User linked to Lab or Award.
This property can be used on the Lab item view page as well as the Award view where currently a standard non-admin user sees 'No view permission' (or current N/A) on the Lab view in the home page redesign branch.

Includes unit tests.  

As I am anxious to get the home page updates into master I ask @utku-ozturk if he might want to merge this branch into the home page update branch or if this should remain separate - if so hope it won't hold things up in getting that branch into master.  

I've included wranglers as reviewers as well as Utku. 

If it is decided to merge this branch independently I will bump the version and add to the changelog.

### DIFF
--- a/src/encoded/tests/data/inserts/lab.json
+++ b/src/encoded/tests/data/inserts/lab.json
@@ -6,6 +6,7 @@
         "address2": "10 Schattuck Street",
         "address1": "Biomedical Bioinfomatics",
         "city": "Boston",
+        "pi": "986b362f-4eb6-4a9c-8173-3ab267227e3d",
         "country": "USA",
         "fax": "000-000-0000",
         "institute_label": "HMS",
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,14 @@ fourfront
 Change Log
 ----------
 
+4.7.0
+======
+
+`PR pi_name calc prop <https://github.com/4dn-dcic/fourfront/pull/1746>`
+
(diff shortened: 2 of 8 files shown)


utku-ozturk (approved):
@aschroed Either way - merging it into the `master` or the home page branch - works for us. Let's do it separately.

src/encoded/tests/data/inserts/lab.json
```diff
@@ -6,6 +6,7 @@
         "address2": "10 Schattuck Street",
         "address1": "Biomedical Bioinfomatics",
         "city": "Boston",
+        "pi": "986b362f-4eb6-4a9c-8173-3ab267227e3d",
```
andreacosolo, line 9: Both this "4DN member" user and the "Test-4DN" lab have status `deleted` on the data portal. Is this OK?

andreacosolo (approved):
Looks good

src/encoded/tests/data/inserts/lab.json
aschroed: I think this is just a leftover of when some inserts were accidentally loaded into the portal.  I don't think it should be a problem.

RahiNav: Looks good!

RahiNav (approved):

PR #1752, opened 2022-12-01
ChIP-seq QC V2 schema and calc props

clarabakker:
Created to handle and display quality metrics generated by the 4DN adoption of the ENCODE ChIP-seq Pipeline.


aschroed (commented):
I've created a front end [trello ticket](https://trello.com/c/tx3F585z/723-qc-metric-overview-view) describing the issues with the overview of QC items from this workflow.  Let's see if they have any changes required so you may want to wait for that feedback just in case changes are required.

utku-ozturk:
@clarabakker FYI, I recently pushed a commit that fixes UI issues addressed in [Trello ticket](https://trello.com/c/tx3F585z/723-qc-metric-overview-view).


03cad3b9a5fbaa18dc68da55cb7c7508
b832edd76cf522508a7091a77b366888

src/encoded/schemas/quality_metric_chipseq_v2.json
```diff
@@ -0,0 +1,1362 @@
...
+                                    "title": "Relative strand cross-correlation coefficient",
+                                    "type": "number",
+                                    "qc_order": 9
+                                }
+                            }
+                        }
+                    }
+                }
+            },
+            "jsd": {
```
utku-ozturk, line 1257: @clarabakker Could you check  `jsd` property since its location looks not clear to me. Is it a top-level QC field like `align_enrich` / `lib_complexity`, or is it a sub-property of `align_enrich`? If it is sub-property, it might be moved into the `properties`.

clarabakker: This all looks great--thank you, Utku! You are right, jsd should be a property of of `align_enrich`.

src/encoded/types/quality_metric.py
```diff
@@ -330,7 +373,6 @@ class QualityMetricChipseq(QualityMetric):
     def quality_metric_summary(self, request):
         return get_chipseq_atacseq_qc_summary(self.properties, 'QualityMetricChipseq')
 
-
```
andreacosolo, line 333: Why deleting this line?

src/encoded/types/quality_metric.py
```diff
@@ -313,6 +313,49 @@ def tooltip(numVal):
...
+            
+            qc_summary.append({"title": "Nonredundant Read Fraction (NRF)",
+                               "value": str(round2(qc.get("lib_complexity")["lib_complexity"]["rep1"]["NRF"])),
+                               "tooltip": "distinct non-mito read pairs / total non-mito read pairs",
+                               "numberType": "float"})
+
+        if 'align' in qc:
+            
+            final_reads = qc.get(
+                "align")["nodup_samstat"]["rep1"]["read1"]  # PE
```
andreacosolo, line 349: If align is not present in qc, then `qc.get("align")` returns None and the following superscript ["nodup_samstat"] etc makes no sense and returns an error.
Either you are sure that "align" is always found (and you can just do `qc["align"]["nodup_samstat"]`), or you need to make it so it does not throw an error.

I see that you used the same style in other places in this calcprop, and I think all of those should be fixed.

andreacosolo (commented):
I just looked at the new type and have a comment that needs to be addressed before this can be approved.

Also the version and changelog need to be updated before merging.

src/encoded/types/quality_metric.py
andreacosolo: Oh, sorry, I did not see the line above: `if 'align' in qc`. I guess I'm not sure why you do a get, but it should work.

andreacosolo: I'm still not sure, in case of SE, what happens at this line. Will there not be a "read1"? Or do you always have a "read1" key which can however have a null value?

clarabakker: In both cases of SE and PE, `qc["align"]["nodup_samstat"]["rep1"]["read1"]` exists and has a value. In the case of SE, the value is zero. The QC metric attached to the processed file from the fastq 4DNFIWQXK28E is one example.

src/encoded/types/quality_metric.py
clarabakker: I am happy to replace it. I believe my thinking was that the single space better groups the two versions of the ChIP-seq quality metric. I am a bit worried about the scenario in which someone jumps to V1 first and does not see V2.

I found spacing generally inconsistent around ChIP-seq, but I also thought a single space might be appropriate given the single line separation between QualityMetricAtacseq and QualityMetricChipseq. Maybe that should also just be two lines.

andreacosolo (approved):

PR #1753, opened 2022-12-01
embed microscope_configuration_master in ExperimentSet

andreacosolo:


andreacosolo:
This is actually not needed right now. Closing without merge.

PR #1758, opened 2022-12-12
Change jwtToken Cookie's SameSite Attribute As Lax

utku-ozturk:
**Trello:** https://trello.com/c/AgoySZB1

TLDR; let's set jwtToken's `SameSite` attribute as `Lax` instead of `Strict`.

### About SameSite Attribute

Currently, the `jwtToken` cookie's `SameSite` attribute is set as `Strict`, which is the most restrictive among the available options: `None`, `Lax`, and `Strict`.

As described in this [article](https://web.dev/samesite-cookies-explained/), `SameSite=Strict` cookies are not included within the request even if it is initiated by a top-level navigation where as `SameSite=Lax` cookies are included:

> If you set SameSite to Strict, your cookie will only be sent in a first-party context. In user terms, the cookie will only be sent if the site for the cookie matches the site currently shown in the browser's URL bar.

> When the user is on your site, then the cookie will be sent with the request as expected. However when following a link into your site, say from another site or via an email from a friend, on that initial request the cookie will not be sent.

> That's where SameSite=Lax comes in by allowing the cookie to be sent with these top-level navigations ..... This makes Lax a good choice for cookies affecting the display of the site with Strict being useful for cookies related to actions your user is taking. 

The description above explains why authentication fails when the link in google sheets is clicked.

### Solution

Neither `SameSite=Strict` nor `SameSite=Lax` offers concrete security, but both, of course, are better than `SameSite=None`. Regarding 4DN and CGAP scopes, 4DN looks to have more flexibility compared to the security restrictions implied by CGAP. The `SameSite=Lax` transition can help improve the end-user experience without sacrificing security. (backing by `Secure=true` might be better, but please keep in mind that localhost/hotseat is still not supporting HTTPS.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@ fourfront
 Change Log
 ----------
 
+4.7.3
+======
+
+`PR jwtToken cookie <https://github.com/4dn-dcic/fourfront/pull/1758>`_
+
+* jwtToken cookie's SameSite=Strict attribute changed as SameSite=Lax
+
 4.7.2
 ======
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1951,11 +1951,11 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.20.1",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.20.1.tgz",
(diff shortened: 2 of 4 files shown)


andreacosolo (commented):
I don't think I have sufficient understanding of the underlying security implications to formally approve this PR, but of course I'm happy if this solves the problem I reported.

willronchetti (approved):
Just needs version bump

PR #1766, opened 2023-01-10
Child Popup Window Updates

utku-ozturk:
**Trello:** https://trello.com/c/dJCyoEJ0

More detailed information is available in https://github.com/4dn-dcic/shared-portal-components/pull/201.


Bianca-Morris (approved):
Don't see any problems with this; should probably be good to go if it's something that's needed urgently. If not, is this deployed somewhere? Would just like to test out that back button on a few of the different pages.

PR #1772, opened 2023-01-26
HiGlass Multivec + HiGlass Upgrades

utku-ozturk:
Trello: https://trello.com/c/nBNMlaxK

### DIFF
--- a/src/encoded/static/components/item-pages/HealthView.js
+++ b/src/encoded/static/components/item-pages/HealthView.js
@@ -131,6 +131,10 @@ export default class HealthView extends React.PureComponent {
                 title : "MicroMeta Version",
                 description : "Software version of MicroMeta package being used."
             },
+            'higlass_version': {
+                title : "HiGlass Version",
+                description : "Software version of HiGlass package being used."
+            },
             'vitessce_version': {
                 title : "Vitessce Version",
                 description : "Software version of Vitessce package being used."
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/item-pages/HealthView.js
```diff
@@ -235,18 +239,15 @@ export default class HealthView extends React.PureComponent {
...
             [ microMetaVersionUsed ] = microMetaFrom.split('#').splice(-1);
         }
-        //extend context to include vitessce version
-        const { dependencies } = installedPackageLockJson;
-        let vitessceVersion;
-        if (dependencies && dependencies.vitessce) {
-            vitessceVersion = dependencies.vitessce.version;
-        }
+        //extend context to include higlass & vitessce version
+        const { dependencies: { 'higlass': { version: higlassVersion } = {}, 'vitessce': { version: vitessceVersion } } } = installedPackageLockJson;
```
Bianca-Morris, line 243: Nice cleanup. I think we might do something like this for CGAP's health page, too, when I do this update.

Bianca-Morris (approved):
JS Code looks good to me. I did take a look at hotseat, but think @alexander-veit or @aschroed would probably know better if things are still displaying/working as they should. Just not familiar enough with this vis to really say either way.

aschroed (approved):
I've looked on hotseat and looks good to me.  Not sure if there are any multivec files there to check so did not check that.

Bianca-Morris:
> I've looked on hotseat and looks good to me. Not sure if there are any multivec files there to check so did not check that.

It's probably a good idea to double check this. On CGAP's last Higlass upgrade, we uncovered a lot of Higlass issues that we thought were simply data-related after testing on an env with more data. If there's an easy/not-super-complicated way to do this, it might be worth it.

aschroed:
> 

I found a multivec file on hotseat and looks good!
http://fourfront-hotseat-1650461284.us-east-1.elb.amazonaws.com/files-processed/4DNFINGT4SXC

PR #1777, opened 2023-02-06
Cypress tests fail due to change in search result's total count

utku-ozturk:
GA Cypress runs started to timeout after the ES7 migration. A new "feature" in ES7 restricts the result's total count to 10000 instead of its exact count. Because of this change, 1. Cypress tests that depend on the result's total count fail, 2. end-users get a total count that looks less precise.

One workaround would be displaying "10.000+ results ", instead of "10.000 results", whereas another is getting the exact count from facet terms where `type=Item`'s `doc_count` is available. In this PR, we preferred the latter approach that only requires changes in `search.py`, since the former affects numerous FF/SPC components and Cypress tests.

**Notes:**

- ~~It is available in hotseat~~
- `dcicsnovault` ~~is currently in beta~~ 7.1.3 installed
- the solution is not working for `frame=raw` queries since facets are not included.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@ fourfront
 Change Log
 ----------
 
+5.0.4
+======
+
+`PR Cypress tests fail due to change in search result's total count  <https://github.com/4dn-dcic/fourfront/pull/1777>`_
+
+* Gets search results' exact count from facet terms where type=Item's doc_count is available
+
 5.0.3
 ======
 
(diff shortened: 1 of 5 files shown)


aschroed (approved):
This looks good to me.  It's too bad that the frame=raw view cannot show the actual total but maybe there isn't much that can be done - @willronchetti any obvious back end solutions?  Not that we have transparent documentation on the available frames, especially for search views so it seems unlikely that this would be an issue.

willronchetti (approved):
Just needs version bump

PR #1780, opened 2023-02-13
update experiment_type.json file to include cfde to experiment name mapping

RahiNav:
Refers to ticket https://trello.com/c/GnF4QRPS/178-add-cfdeterm-field-to-experimenttype-schema-and-add-values

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,11 @@ fourfront
 Change Log
 ----------
 
+5.0.3
+======
+
+* Updates to experiment_type.json file to include cfde term based on the experiment name to obi mapping.
+
 5.0.2
 ======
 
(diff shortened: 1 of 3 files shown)


aschroed (changes requested):
Hi Rahi - adding the full mapping somehow to the schema is not intention here.  What we want is to add a new property to the schema that will be very similar to the existing 'controlled_term' property.
`        "controlled_term": {
            "title": "Ontology Term",
            "description": "Ontology term that matches this experiment type",
            "type": "string",
            "linkTo": "OntologyTerm"
        }
`
that is designed as a linkTo an OntologyTerm item that exists as an item in the portal (that comes from the OBI ontology).  Once this property is added to the schema then the next task would be to make the link between the ExperimentType item and the OBI term Item in the portal by patching that new property with the uuid (or term_id) of the existing OBI term.

RahiNav:
I see. I have updated the file again to complete the first step.

aschroed (approved):

PR #1785, opened 2023-02-28
added cdfe term to associated exp type

RahiNav:
Added appropriate CFDE terms to the corresponding experiment type.

Note: 
Certain labels in CFDE and OBI slightly differ, e.g. 
In CFDE: 
"TSA-seq": "tyramide signal amplification sequencing assay (provisional term)"    

In OBI:
"TSA-seq": "tyramide signal amplification sequencing assay"

I have used the exact labels given in CFDE portal (https://app.nih-cfde.org/chaise/recordset/#1/CFDE:assay_type@sort(nid))

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ fourfront
 Change Log
 ----------
 
+5.1.3
+======
+
+* Added corresponding 'cfde term' to the experiment type
+
+
 5.1.2
 ======
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 # Note: Various modules refer to this system as "encoded", not "fourfront".
 name = "encoded"
-version = "5.1.2"
+version = "5.1.3"
 description = "4DN-DCIC Fourfront"
 authors = ["4DN-DCIC Team <support@4dnucleome.org>"]
 license = "MIT"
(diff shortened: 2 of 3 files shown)


aschroed (changes requested):
It's good that you are adding these terms to the master inserts - however, unfortunately it will not work the way you have done this for a couple of reasons.
1. All the OntologyTerm items do not get loaded when deploying local and that is the main place that these inserts are used.  In order to have this work you would need to add inserts to master_inserts for all the relevant OntologyTerms but I do not recommend that you do this.  Instead I recommend removing the cfde_term property and values from the inserts.
2. Even if you were to add the OntologyTerm items to the inserts you need to refer to those term by either uuid or term id and not term name.  This is because you need to use either the uuid or the name_key (which is a property defined in the types file for an item) to refer to a linkTo item.  If you want to go over how this works let me know.  
3.   Since these are linkTo values even if you could have used the name the name would need to match that of an existing OntologyTerm in the database.  The CFDE names are not those of the official OBI terms.  The '(provisional term)' part of the term was that used until the term officially was added to the ontology.  The OBI term in the portal lacks that part of the name because it is no longer provisional and is now official.

Basically this branch is not needed and the task is to patch the ExperimentType items on the portal with the cfde_term field with the values as OBI terms that are items in the portal.  Let me know if you want to discuss.

PR #1790, opened 2023-03-08
DRAFT - Upd ont gen

aschroed:
-  Added an option `--phase` to phase the generated json so that new terms that need to be POSTed required fields are written to the file first before terms that are updates and the non-required fields for the newly posted terms are PATCHed.

In theory this should not be necessary if the entire file with all the updates can be loaded in a single process as load_xl does the phasing for us, however, if the file needs to be chunked or the process is interrupted before all the terms have gone through the load_xl first phase of posting required fields due to all the interdependencies of OntologyTerm items - each term has linkTo parents and slim_term OntologyTerm - that need to exist before PATCHES can be successful.

- Added  an option `--owlfile` to allow you to provide a local ontology OWL file to the script - rather than relying on file download from the value in the `download_url` property of the Ontology item

- bug fix to properly handle updates to terms that belong to more ontologies if a term is deprecated/obsoleted in one of them.  This feature needs to be better tested if processing multiple ontologies at once (though the standard process has been to only update one ontology at a time)


src/encoded/commands/generate_ontology.py
```diff
@@ -1053,12 +1101,12 @@ def main():
...
             if not v and ontology.get('ontology_name').upper() == 'UBERON':
                 try:
-                    result = requests.get('http://svn.code.sf.net/p/obo/svn/uberon/releases/')
-                    release = result._content.decode('utf-8').split('</li>\n  <li>')[-1]
-                    v = release[release.index('>') + 1: release.index('</a>')].rstrip('/')
+                    result = requests.get('https://api.github.com/repos/obophenotype/uberon/git/refs/tags')
+                    v = result.json()[-1].get('ref').split('/v')[-1]
                 except Exception:
                     print('Unable to fetch Uberon version')
             if v and v != ontology.get('current_ontology_version', ''):
```
dmichaels-harvard, line 1112: I think that to respect the --full argument we want to change this line (1064) to:
```if (v and v != ontology.get('current_ontology_version', '')) or args.full:

dmichaels-harvard: And also, to fix a bug I was running into (running localling anyways), the very next line after that should initialize `prev` to `[]`, i.e. ```prev = []```

aschroed: Absolutely agree on initializing prev.
In regard to the --full argument I had not considered the use case of reverting something to an earlier version of the ontology and I am not sure that if one tried to do so things would work exactly as hoped - which I think is the use case you are considering?  The --full option was intended as more of a checking mechanism where a file was generated that contained both terms to be posted or updated as well as existing unchanged terms that are usually filtered out so the file that was generated was 'complete' in reference to the db it was run against - the usual filtering reduces the file size and avoids unnecessary patches but sometimes it helps to see everything that exists in the ontology obtained from the maintainers if troubleshooting is required.  And generally in that case you don't need the ontology version info but I don't think it hurts.

dmichaels-harvard: Okay yeah, I can take that out. I was just getting confused when generating an ontology file and wasn't getting any ontology (as opposed to term) records, so I figured the --full was intended to do that; but I didn't realize the fuller implications of that.

PR #1796, opened 2023-04-26
from test_file.py removed unused imports and marked several tests as integrated

aschroed:


src/encoded/tests/test_file.py
```diff
@@ -4,14 +4,12 @@
...
 from dcicutils.beanstalk_utils import source_beanstalk_env_vars
-from dcicutils.misc_utils import file_contents
-from dcicutils.qa_utils import MockBoto3, MockBotoS3Client
 from pyramid.httpexceptions import HTTPForbidden
 from unittest import mock
 from ..types.file import FileFastq, post_upload, external_creds
 
 
-pytestmark = [pytest.mark.setone, pytest.mark.working]
+pytestmark = [pytest.mark.setone, pytest.mark.working, pytest.mark.integrated]
```
netsettler, line 12: Marking integrated here means everything in this file is an integrated test. I don't think you want that. I'd remove this marker and leave the others on individual functions.

netsettler (approved):
There's a trivial change I think you should made, and if you're OK with that, I think this is fine. Thanks.

src/encoded/tests/test_file.py
aschroed: Removed and added an inline comment. Will merge when CI completes.

PR #1797, opened 2023-05-04
Facet Terms Grouping

utku-ozturk:
Trello: https://trello.com/c/tRSyktQ5

Demo: http://fourfront-hotseat-1650461284.us-east-1.elb.amazonaws.com/browse/?experimentset_type=replicate&type=ExperimentSetReplicate

Please also refer to: https://github.com/4dn-dcic/shared-portal-components/pull/231

The update allows for a hierarchical display of terms grouped under a grouping term. It also enables searching and selection by group names and individual terms.

You must add the `group_by_field` property in schema json to define grouping for a facet. The experiment type facet's terms of ExperimentSet are grouped in this PR.

    "facets": {
       ...
        "experiments_in_set.experiment_type.display_title": {
           ...
            "group_by_field": "experiments_in_set.experiment_type.assay_subclass_short",
        },

Screen Shot 2023-05-04 at 11 57 48


src/encoded/search.py
```diff
@@ -1322,13 +1323,31 @@ def set_facets(search, facets, search_filters, string_query, request, doc_types,
         else:
 
             facet['aggregation_type'] = 'terms'
-            term_aggregation = {
-                "terms" : {
-                    'size'    : 100,            # Maximum terms returned (default=10); see https://github.com/10up/ElasticPress/wiki/Working-with-Aggregations
-                    'field'   : query_field,
-                    'missing' : facet.get("missing_value_replacement", "No value")
+            
+            term_aggregation = None
```
willronchetti, line 1327: I don't think you need this declaration since `term_aggregation` is set unconditionally in the if/else below

willronchetti (approved):
Back-end looks like it will work, one thing I will mention though is you may want to test the behavior when someone sets the group_by_field in the schema but doesn't have a corresponding embed for that field. It may even be useful to add a check for this to schema validation. This would need to be done in snovault though so may not be worth doing at this time.

Bianca-Morris (approved):

PR #1804, opened 2023-05-17
Cypress Test for Not Facet

utku-ozturk:
Trello: https://trello.com/c/qIw1omhJ

New Cypress test added for /browse page:

- it toggles between included and excluded properties in facets
- it excludes a `award.project` term and compares the exact Exp Set’s before and after counts
- it removes the excluded item's selection, then includes it, and compares the before and after counts.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,16 @@ fourfront
 Change Log
 ----------
 
+5.3.5
+=====
+
+`PR Cypress test for not facet  <https://github.com/4dn-dcic/fourfront/pull/1804>`_
+
+* it toggles between included and excluded properties in facets
+* it excludes a award.project term and compares the exact Exp Set’s before and after counts
+* it removes the excluded item's selection, then includes it, and compares the before and after counts.
+
+
 5.3.4
 =====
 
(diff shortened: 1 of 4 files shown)


Bianca-Morris (commented):
Didn't run this locally. It would be good to try to figure out a way to get rid of those `.wait()`s. It makes the tests a [little flakier and take longer to run](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting). Could maybe use .should() with a callback function to wait for a change in length/number of search results. Also could maybe do an intercept on the request, although I'm not sure how difficult that would be.

utku-ozturk:
@Bianca-Morris Yes, you are absolutely right. Even if we try to avoid using wait(), we have already used it after click() events. This is a sample `wait` to `should` conversion:

**version 1 (using wait())**

```
cy.get('.my-button').click().wait(3000); // Waiting for 3 seconds
```

**version 2(using should())**

```
cy.get('.my-button').click();
cy.contains('.new-page-element', 'Welcome').should('be.visible');
```

I am not sure it is applicable to all cases; we can do our best to convert all existing wait statements. If makes sense, let's track it in a separate ticket.

PR #1818, opened 2023-06-26
Small tweak to generate-ontology script

aschroed:
To account for the search limit of 100K items broke up the single query for all existing ontology terms into multiple queries (by ontology) and results to dictionary to avoid terms that are part of multiple ontologies being redundant in the returned result.

### DIFF
--- a/src/encoded/commands/generate_ontology.py
+++ b/src/encoded/commands/generate_ontology.py
@@ -390,17 +390,21 @@ def get_slim_terms(connection):
     return slim_terms
 
 
-def get_existing_ontology_terms(connection):  # , ontologies=None):
+def get_existing_ontology_terms(connection, ontologies=None):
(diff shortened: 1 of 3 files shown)


dmichaels-harvard (approved):
LGTM

src/encoded/commands/generate_ontology.py
```diff
@@ -997,6 +1001,11 @@ def write_outfile(to_write, filename, pretty=False):
         else:
             json.dump(to_write, outfile)
 
+def filter_ontologies(ontologies):
+    for i, o in enumerate(ontologies):
+        if o['ontology_name'].startswith('4DN') or o['ontology_name'].startswith('CGAP'):
```
willronchetti, line 1006: Do we really care about CGAP here?

willronchetti (approved):

PR #1819, opened 2023-06-28
UA to GA4 Migration

utku-ozturk:
https://trello.com/c/LOfJamzG

SPC PR: https://github.com/4dn-dcic/shared-portal-components/pull/233


willronchetti (commented):
I don't see any backend changes here necessary to utilize the new secret when submitting events - generally though I see no issue with the UI changes

src/encoded/static/ga_config.json
```diff
@@ -1,25 +1,13 @@
...
-        "downloads"         : 2
+        "data.4dnucleome.org" : ["GTM-KZ9JZM2", "G-23BW5XEWGZ"],
+        "mastertest.4dnucleome.org" : ["GTM-3DT6W3Z", "G-KW98FB0NDW"],
+        "staging.4dnucleome.org" : ["GTM-KZ9JZM2", "G-23BW5XEWGZ"],
+        "fourfront-hotseat-1650461284.us-east-1.elb.amazonaws.com" : ["GTM-3DT6W3Z", "G-KW98FB0NDW"],
+        "hotseat.4dnucleome.org" : ["GTM-3DT6W3Z", "G-KW98FB0NDW"],
+        "localhost:8000" : ["GTM-WMSVXV3", "G-JWY59WYF4N"],
+        "localhost:6543" : ["GTM-WMSVXV3", "G-JWY59WYF4N"],
+        "default" : []
     },
```
Bianca-Morris, line 11: Can you remind me how the `dimensionNameMap` and the `metricNameMap` used to work, and why they are no longer necessary?

src/encoded/types/file.py
```diff
@@ -1459,98 +1459,74 @@ def update_google_analytics(context, request, ga_config, filename, file_size_dow
...
-        # Product Category from @type, e.g. "File/FileProcessed"
-        "pr1ca": "/".join([ty for ty in reversed(context.jsonld_type()[:-1])]),
-        # Product "Variant" (supposed to be like black, gray, etc), we repurpose for filetype for reporting
-        "pr1va": file_type
-        # "other" MATCHES THAT IN `file_type_detaild` calc property,
-        # since file_type_detailed is used on frontend when performing "Select All" files.
+        "client_id": ga_cid,
+        "timestamp_micros": str(int(datetime.datetime.now().timestamp() * 1000000)),
+        "non_personalized_ads": False,
+        # "user_agent": request.user_agent,
```
Bianca-Morris, line 1486: Is this not necessary?

src/encoded/static/components/browse/components/SelectedFilesController.js
```diff
@@ -155,9 +160,13 @@ export class SelectedFilesController extends React.PureComponent {
             logger.error("Supplied accessionTriple is not a string or array of strings/arrays:", accessionTriple);
             throw new Error("Supplied accessionTriple is not a string or array of strings/arrays:", accessionTriple);
         }
-        const newlyAddedFileItems = [];
+        //it is initialized inside of setState, since setState is called twice
+        //in strict mode and newlyAddedFileItems would have duplicate items
+        //more info: https://github.com/facebook/react/issues/12856#issuecomment-390206425
```
Bianca-Morris, line 165: Thanks for clarifying this. I remember Alex had explained this at one point, but I couldn't remember the exact reason.

Bianca-Morris (approved):
Looks like you updated to address Will's concern. I didn't get to test this (honestly not sure how I would test it; this is something we should make time to discuss at some point -- I don't have a ton of experience with Google Analytics or Tag Manager), but seems to mostly make sense.

I think the only major thing I'm wondering at this point is if there's anywhere that the relationship between GA's cart/product terminology and what those things mean in our app is documented? Is adding an item to the cart, just selecting the item, or does it only count if you actually take some action afterward. Why? You can tell some of this by reading the code, but having a general overview would be nice for people new to the system.

I also think the one-word `// analytics ` comments are mostly useless; not sure if you were planning on adding more to those. I think if it's not going to provide more information about what each analytics event or action is doing, probably better to just get rid of it, since you can tell from the method names that they're related to analytics, and otherwise it just adds clutter. If there are any particularly complex events or calls to those analytics utility methods, would be good to comment them with more detail, though.

Nice job! Thanks for all your work on this, and your patience.

src/encoded/static/ga_config.json
utku-ozturk: @Bianca-Morris Even if UA and GA4 let users define their custom dimensions and metrics, dimensions/metrics cannot be submitted by their name in UA but you need to POST them as dimension 1, dimension 2 ... dimension N. So a mapping between the name and index was required in the old implementation. The restriction was removed in GA4.

src/encoded/types/file.py
utku-ozturk: `client_id `is required to identify the user's session, whereas `timestamp_micros ` determines when the event is triggered.

src/encoded/static/components/browse/components/SelectedFilesController.js
utku-ozturk: Yeah, I've also lost 2 hours determining why it is not working. Dan Abramov had a good explanation in the link.

utku-ozturk:
> I also think the one-word `// analytics ` comments are mostly useless; not sure if you were planning on adding more to those. I think if it's not going to provide more information about what each analytics event or action is doing, probably better to just get rid of it, since you can tell from the method names that they're related to analytics, and otherwise it just adds clutter. If there are any particularly complex events or calls to those analytics utility methods, would be good to comment them with more detail, though.

@Bianca-Morris Yes, absolutely you're right.  I have used //analytics for bookmarking, otherwise it was hard to track all. it makes it easy to list all. I will remove them.

PR #1840, opened 2023-08-25
Updated to pyyaml 6.0 and generate-local-access-key script and remove access_key.py/json

dmichaels-harvard:
* Added generate-local-access-key script (from snovault) to pyproject.toml;
  orignally created for smaht-portal since early in development no way to
  create an access-key normally using the UI; but generall useful/convenient.
* Removed types/access_key.py and schemas/access_key.json as the ones in snovault are sufficient.
* Changed pyyaml version to 6.0.


willronchetti (approved):

CHANGELOG.rst
```diff
@@ -6,6 +6,16 @@ fourfront
 Change Log
 ----------
 
+6.3.0
+======
+
+* Added generate-local-access-key script (from snovault) to pyproject.toml;
+  orignally created for smaht-portal since early in development no way to
+  create an access-key normally using the UI; but generall useful/convenient.
```
netsettler, line 14:
```suggestion
  create an access-key normally using the UI; but generally useful/convenient.
```

src/encoded/commands/extract_test_data.py
```diff
@@ -122,7 +122,7 @@ def anon_pipeline():
 
 
 def run(pipeline, inpath, outpath):
-    for item_type in ORDER:
+    for item_type in ORDER():
```
netsettler, line 125: See [my related comment on your snovault PR](https://github.com/4dn-dcic/snovault/pull/266#discussion_r1328305859), but I found it really weird to see (a) a function here that is all-uppercase, and (b) a change in the semantics of ORDER, which had previously been a list. I think if the semantics changes in so strong a way, it should just have a different name, so that it doesn't look to someone who doesn't know about the name like you've used it wrong.

So my suggestion is to rename the function to something like `load_order_for_app` in the other PR and then use the newly chosen name here.

PR #1852, opened 2023-09-26
Upgrade to Python 3.11

dmichaels-harvard:


willronchetti (commented):
You need to bump the Docker Python version by changing the base image, deploy to mastertest and test out particularly the access key change (generate, ensure 4DN specific fields are there etc)

pyproject.toml
```diff
@@ -32,23 +32,26 @@ classifiers = [
...
+    'Programming Language :: Python :: 3.10',
+    'Programming Language :: Python :: 3.11'
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8.1,<3.10"
-awscli = ">=1.25.36"
-boto3 = "^1.26.133"
-botocore = "^1.27.36"
+python = ">=3.8.1,<3.12"
```
netsettler, line 42: 3.8.1 or 3.8? See my comments on cgap-portal.

pyproject.toml
```diff
@@ -32,23 +32,26 @@ classifiers = [
     # that you indicate whether you support Python 2, Python 3 or both.
     'Programming Language :: Python :: 3',
     'Programming Language :: Python :: 3.7',
```
netsettler, line 34: Remove 3.7 here.

src/encoded/commands/extract_test_data.py
```diff
@@ -122,7 +122,7 @@ def anon_pipeline():
 
 
 def run(pipeline, inpath, outpath):
-    for item_type in ORDER:
+    for item_type in ORDER():
```
netsettler, line 125: I've made comments about this change to ORDER on other PRs. I really don't like this change.

netsettler (commented):
I'm making comments but withholding approval pending the ORDER() issue being addressed.

src/encoded/commands/extract_test_data.py
dmichaels-harvard: Oh shoot, I made that change in snovault, and in smaht-portal, but forgot to propagate to fourfront - will do - thanks.

pyproject.toml
dmichaels-harvard: I didn't change that from master, it used to be: python = ">=3.8.1,<3.10" ... not sure why it would be 3.8.1 ...
(Same comment from cgap-portal ... and Will replied: I think I recall from a long time ago 3.8.1 has some specific issues with cgap-portal IRC hence the explicit exclusion)

pyproject.toml
dmichaels-harvard: Oversight, thanks, done.

willronchetti (approved):
Approving now since integrated testing looks good

PR #1870, opened 2024-02-01
Individual Primate schema updated

RahiNav:
Added https://data.4dnucleome.org/organisms/9544 and https://data.4dnucleome.org/organisms/9483/ uuids to the Individual Primate linkEnum

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ fourfront
 Change Log
 ----------
 
+7.1.5
+=====
+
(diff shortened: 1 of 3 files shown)


aschroed (approved):

aschroed:
I tested this locally and it looks good.  One additional suggestion would be to add an insert for an Individual for at least one if not both of the new organisms in `tests/data/inserts/individual_primate.json` as this I believe is the first case of multiple organisms and therefore no default value in an Individual schema.  However, this is not blocking.

aschroed (approved):
Looks good.

PR #1874, opened 2024-02-28
new organisms schema update

RahiNav:
Added https://data.4dnucleome.org/organisms/9544 and https://data.4dnucleome.org/organisms/9483/ uuids to the Individual Primate linkEnum and added them as inserts in organism.json file.


aschroed (approved):
Looks good.

PR #1878, opened 2024-03-08
Repair Accession Validation

willronchetti:
- Brings in fixes to ensure validation of accessions (and other custom validators) occurs


aschroed (commented):
Still looking but seeing that one of the test insert workflows has failed to load with an invalid accession error: 4DNWFP00U73B and I see that you changed the regex - not sure if that is the issue?

aschroed (approved):

PR #1893, opened 2024-05-28
Fix Cypress 03d Test

utku-ozturk:
- Currently, there is a known difference between QuicknfoBar's file count (raw + processed + of) and  the selected file count after the "Select All" button click (raw + processed). The 03d test was not taking into account this difference, fixed.
- Chart's files aggregation now includes OPF counts.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,15 @@ fourfront
 Change Log
 ----------
 
+7.5.6
+=====
+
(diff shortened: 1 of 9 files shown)


utku-ozturk:
[The "Select All" button not turns into "Deselect All" in /browse](https://trello.com/c/p409yflj/280-the-select-all-button-not-turns-into-deselect-all-in-browse)

PR #1895, opened 2024-06-06
Fix DRS

willronchetti:
- Update DRS downloads URLs to direct to open data where applicable
- TODO bring in snovault version that forces `application/json` for DRS


dmichaels-harvard (approved):

aschroed (commented):
Can you add open_data_url in fixture in a test?

PR #1900, opened 2024-07-18
Ajs upd wf schema

aschroed:
* Added 'tags' mixing to workflow schema
* Added 'max_runtime' property to workflow schema
* Added 'current pipeline' and 'accepted_pipelines' to experiment_type for the names
  of completed pipelines used by foursight


src/encoded/schemas/experiment_type.json
```diff
@@ -246,6 +247,20 @@
             "description": "CFDE term that corresponds to the 4DN experiment type based on OBI ontology",
             "type": "string",
             "linkTo": "OntologyTerm"
+        },
+        "current_pipeline": {
```
clarabakker, line 251: How do you envision this property working for pipelines that are run on many different experiment types (or similarly, experiment types with several kinds of pipelines)? E.g. QC pipelines or format conversion. Maybe an array would work, although accepted_pipelines would also need adjusting to clarify which versions go with which pipelines?

clarabakker: The QC and format conversion pipelines will also search for files rather than experiment sets, so workflow information could be stored on the file level instead. I'm not sure if that is reasonable or not. As long as it's okay to pull the experiment type info from the files to get the current and accepted pipelines, putting everything in the experiment type is probably okay.

clarabakker: I also may be completely misunderstanding how this system will work, so I apologize if I have!

aschroed: The purpose of these changes is to avoid having to maintain sometimes redundant hard coded info when it could be maintained in the db allowing easier updates as needed and get rid of the redundancy.  I started with the info that was redundant in foursight and dcicwrangling and is definitely used by both.  There is more redundant info that I really don't think is used in dcicwrangling any more.  There is likely more that can be factored out but it's a cost benefit issue.

The fields on the ExperimentType schema are corresponding to those currently [here](https://github.com/4dn-dcic/foursight/blob/4a16a920534776fcb40362cbb69e520c79bcd31f/chalicelib_fourfront/checks/helpers/wfr_utils.py#L160) in foursight, which I had asked about earlier and didn't think they were used for QCs etc. but were primarily for marking sets with the completed pipelines.  Are they used for other purposes?  

The tags on the workflows could be either 'accepted' or 'current' and this is what would be used to generate the dictionary [here](https://github.com/4dn-dcic/foursight/blob/4a16a920534776fcb40362cbb69e520c79bcd31f/chalicelib_fourfront/checks/helpers/wfr_utils.py#L18).  These are the QC relevant ones I believe but each workflow will have it's own tag so should not be an issue.

clarabakker: Yes, good point, I am getting the usage of tags a little mixed up. Hi-C experiments do receive additional tags for the compartments caller and insulation score caller runs, but those are the only ones and could continue using their separate dictionary.

The accepted_versions dictionary is used to automatically generate some queries and to patch newly processed sets, so I was concerned a pipeline (e.g. compartments) would complete running and use the new experiment type field (or dictionary based off of the field) to mistakenly patch the Hi-C Pipeline tag instead. However, this is more of a special case rather than typical, so it would be okay to leave it.

clarabakker (approved):
Okay, seems good!

PR #1903, opened 2024-07-25
Ajs rb wfschema v

aschroed:
Reverted workflow.json schema version back to 7 (from 8) as an upgrader was not added and schema changes were just addition of new properties so upgrade was not truly needed.


utku-ozturk:
[react-bootstrap v2 upgrade](https://trello.com/c/Y77EYoZP/331-react-bootstrap-v2-upgrade)

PR #1907, opened 2024-09-09
new ignored enum for procfile

RahiNav:
Added "fragments" as one of the ignored enum in processed file schema (this is for Nathan's datasets).


aschroed (commented):
For this case I might add to the suggested_enum list rather than ignored_enums?  Could be generally useful value for other types of files.

aschroed (approved):

PR #1908, opened 2024-09-10
new expset property

RahiNav:
Added new property "disease_name" to experiment_set schema. It will use the ontology term already in 4DN most likely from MONDO (via EFO).

### DIFF
--- a/src/encoded/schemas/experiment_set.json
+++ b/src/encoded/schemas/experiment_set.json
@@ -97,12 +97,17 @@
             "lookup": 20,
             "formInput": "textarea"
         },
-        "disease_name": {
+        "disease_term": {
             "title": "Disease Name",
-            "description": "Ontology term for the disease name associated with the set.",
-            "type":  "string",
+            "description": "Ontology term for the disorder or disease name associated with the set.",
+            "type":  "array",
             "lookup": 40,
-            "linkTo": "OntologyTerm"
+            "items":{
+                "title": "Disorder/Disease Name",
+                "description": "Disorder/Disease ontology term(s)",
+                "type": "string",
+                "linkTo": "OntologyTerm"
+            }
         },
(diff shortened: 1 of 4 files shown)


src/encoded/schemas/experiment_set.json
```diff
@@ -97,6 +97,13 @@
             "lookup": 20,
             "formInput": "textarea"
         },
+        "disease_name": {
```
aschroed, line 100: Minor suggestion to call this disease_term - helps convey it is a link to an OntologyTerm.

src/encoded/schemas/experiment_set.json
```diff
@@ -97,6 +97,13 @@
...
+            "title": "Disease Name",
+            "description": "Ontology term for the disease name associated with the set.",
+            "type":  "string",
```
aschroed, line 103: I'd suggest making this property an array field so more than one disease/disorder term can be added.

src/encoded/schemas/experiment_set.json
```diff
@@ -97,6 +97,13 @@
             "lookup": 20,
             "formInput": "textarea"
         },
+        "disease_name": {
+            "title": "Disease Name",
+            "description": "Ontology term for the disease name associated with the set.",
```
aschroed, line 102: I think the description is OK but maybe add 'disorder' as not everything is a 'disease' i.e. for the disease or disorder ...

aschroed (commented):
I left a couple of comments that I think merit consideration and potential changes so I will wait to officially approve until you consider.

Also you might as well add an example to the 'tests/data/inserts/experiment_set.json' - you don't need to create a new item just add this property and an approriate term to an exising item.

aschroed (approved):
Looks good!

PR #1915, opened 2024-10-28
Update treatment item display title

aschroed:
* Add 'override_treatment_title' property to Treatment items
* Tweak the display_title calcprop to:
  - use the override_treatment_title prop if present
  - for biological treatments with constructs but no biological agent use construct names in treatment display_title
* added tests

### DIFF
--- a/src/encoded/types/treatment.py
+++ b/src/encoded/types/treatment.py
@@ -95,6 +95,7 @@ class TreatmentAgent(Treatment):
             )
             disp_title = "Transient transfection of " + plasmids + conditions
         elif treatment_type == "Biological":
+            disp_title = ''
             if biological_agent or constructs:
                 if biological_agent:
                     disp_title = biological_agent + " treatment" + conditions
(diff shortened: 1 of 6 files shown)


src/encoded/types/treatment.py
```diff
@@ -91,8 +94,24 @@ def display_title(self, request, treatment_type=None, chemical=None,
...
+                    const_str = ''
+                    suffix = ''
+                    if len(constructs) > 3:
+                        suffix = ' and {} more'.format(len(constructs) - 3)
+                        constructs = constructs[0:3]
+                    const_str = ", ".join(
+                        [get_item_or_none(request, construct).get('display_title') for construct in constructs]
+                    )
+                    const_str = const_str.rstrip(", ")
+                    disp_title = const_str + suffix + " treatment" + conditions
```
clarabakker, line 112: Maybe "treatment(s)" or a conditional to determine if it's one or more?

aschroed: My argument here is that even if there are more than one construct used in the treatment it is still a single treatment.  If it's relevant that multiple separate treatment regimens were employed then multiple Treatment items should be entered in the 'treatments' field and then multiple display titles, each with its specific set of agents and conditions will be listed.

PR #1918, opened 2024-11-19
Fixup some inserts data to make compatible with what is in production.

dmichaels-harvard:


aschroed (approved):
I think this is fine and I was able to build local successfully.

However, other than the award uuid update in master-inserts I'm not sure why the items in inserts needed to be updated?

PR #1936, opened 2026-07-28
Align Fourfront with snovault 11.35.2 / SMaHT: dep bump, search & batch-download efficiency + security, test & deploy hardening

willronchetti:
## Summary

Fourfront-only alignment fixes from the comparative audits against `dcicsnovault` 11.35.2 and SMaHT Portal, plus a live S3 credential bug found along the way. Product-specific schemas, facets, auth policy, Beanstalk/deploy behavior and the local `search.py`/`batch_download.py` forks are preserved. No SMaHT/Snovault/infra repos were touched.

## What's in it

**1. Dependency bump — `dcicsnovault` 11.27.0 → 11.35.2, `dcicutils` 8.18.3 → 8.18.8.** Picks up snovault #335, a live stale-Elasticsearch-document bug on Fourfront's array-of-object `linkTo` embeds, plus indexing/ES efficiency work and attachment/JWT/AccessKey security fixes. All 133 snovault symbols Fourfront uses resolve unchanged; the lock delta is the two bumps plus a dropped forecasting sub-tree nothing imports.

**2. File upload & download credentials — `sts:GetFederationToken` → `sts:AssumeRole`.** *A live outage risk, not a cleanup.* `external_creds()` minted temporary S3 credentials with `GetFederationToken`, which AWS accepts only from an IAM user holding long-lived access keys and rejects outright from any caller that itself holds temporary credentials — i.e. our ECS task role, and the GitHub Actions OIDC role adopted in 8.9.2. It works today only because static access keys still sit in the application-configuration secret. The moment those go, users can neither upload files nor download them: `@@download` re-derives credentials for any file without a stored `external` propsheet. Both paths run through the same function, so one change fixes both. This is the migration `encoded-core` 1.0.0 and cgap-portal already shipped. Credentials remain scoped to `s3:PutObject` on the single target key — no privilege change, no static credentials, no other token flow touched.

**3. Search efficiency** (ports of snovault #318 into the local fork): drop aggregations and total-hit tracking on later pages of a `limit=all` scan, stop pulling `embedded.*` for object/raw frames, skip facet aggregations the response discards. `frame=embedded` behavior is unchanged.

**4. `batch_download` hardening:** neutralize CSV/TSV formula injection (CWE-1236) across all three exports, and bound `/report.tsv` to the rendered columns instead of fetching every embedded field.

**5. Test/CI reliability:** the `workbook` fixture now waits for DB/ES counts to agree instead of hiding flakiness behind `--force-flaky`; deprecated `@pytest.yield_fixture` removed (pytest 8); per-run SQS indexer queues cleaned up; static guardrail against shipping `pyramid.debug_authorization`.

**6. Docker / deploy hardening:** nginx supervised under supervisord with a fail-fast entrypoint, bounded upstream retries, an `nginx -t` build gate, pinned OpenSearch image and `configure-aws-credentials@v4`.

## ⚠️ Required deployment configuration

`S3_UPLOAD_ROLE_ARN` must be provisioned — into the application-configuration identity for deployed environments, or the environment for CI and local runs — pointing at a role the portal can assume with `s3:PutObject` on the upload and wfout buckets.

It is deliberately **not** added to `.github/workflows/main.yml`: this repo has no OIDC-role secret to reference, and whether the existing deployment role permits self-assumption can't be verified from here. **Until it is provisioned, `test_file.py`'s upload-credential assertions will fail** — they mint credentials for real and are green today only via the mechanism being retired.

## Testing

Four ES-free test modules, **69 tests, all passing offline**: query shape, formula-injection neutralization, the `debug_authorization` guardrail, and 15 new tests in `test_file_upload_credentials.py` pinning the credential mechanism (AssumeRole used on both the upload and the download path, no access keys reach `boto3`, policy still scoped to the one key, failures still degrade instead of 500ing). Reverting the credential change fails 7 of the 15.

Also verified offline: `poetry check --lock`, `nginx -t` on the committed config, CI YAML parse.

**Not runnable offline** (no AWS credentials, Elasticsearch, Postgres or Docker daemon here): the full `workbook`, indexing and security suites, and `docker build`. CI on this PR must cover those.

## Deferred, with reasons

- **Multi-stage Dockerfile + BuildKit/buildspec rewrite** — needs real `docker build` validation.
- **Staggered/selective reindex on deploy** — needs `loadxl_order()` reconciled with `ITEM_INDEX_ORDER` first, or it silently under-reindexes.
- **`clear-db-es-contents` / `memlimit` fork retargeting to snovault** — needs Beanstalk-preserving tests.
- **Snovault #330** (secondary-index PG coalescing) — unreleased beta.
- **Retiring the local `search.py`/`batch_download.py` forks** — biggest lever, highest regression risk against product facets; separate follow-up.
- **`read_single_sheet` breakage** in the `extract-test-data`/`import-data` dev CLIs — pre-existing, not bump-caused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aschroed (commented):
Trying to deploy local and seeing this issue with 'document' insert.
webtest.app.AppError: Bad response: 422 Unprocessable Entity (not 200 OK or 3xx redirect for http://localhost/dcf15d5e-40aa-43bc-b81c-32c70c9afb48)
b'{"@type": ["ValidationFailure", "Error"], "status": "error", "code": 422, "title": "Unprocessable Entity", "description": "Failed validation", "errors": [{"location": "body", "name": ["attachment", "href"], "description": "Incorrect file type. (Appears to be application/octet-stream)"}]}'

